Issue #4, opened 2021-06-02
Configure deployment automation


malparty:
## Why?

Ensure no manual action is made for deployment, thus reducing the risks for human errors.
Increase traceability of events while enforce PR & validations before any deployment.

### Additional notes:
Define which hosting provider to use (Heroku, AWS, Google Cloud or Digital Ocean)
Configure auto-deploy for 2 environments:

- Staging: `develop` branch
- Production: `main` branch

## Who Benefits?
Project team & stakeholders

junan:
> Staging: release/* branches
 
 @malparty The `develop` branch should be used for staging. For staging we don't create any Release PR. Once we merge a PR into `develop` branch, it should be deployed to staging server automatically(via github action)

malparty:
Well noted, task description updated accordingly, thanks! :)

PR #32, opened 2021-06-09
#4 Configure deployment automation

malparty:
[#4 Configure deployment automation](https://github.com/malparty/google-search-ruby/issues/4)
## What happened

GitHub Action `Deploy Heroku` is now working for staging env and deploys the app [here](https://google-search-ruby-staging.herokuapp.com/)

## Insight

- Update deploy_heroku.yml to map with the right registry token and Heroku App name (underscores not accepted by heroku)
- Create default route so that the test can be fully done

## Proof Of Work
[Staging App available here](https://google-search-ruby-staging.herokuapp.com/)

image

image


malparty:
(Unit tests are passing, I just did not bring the changes from previous PR into this branch)

config/routes.rb
```diff
@@ -1,3 +1,5 @@
 Rails.application.routes.draw do
+  root to: 'keyword#index'
+  get 'keyword/index'
```
ankitkalia1195: can we use `resources :keywords, only: :index`

malparty: Yes thanks!
I've also renamed keywordController into keyword**_s_**Controller, it seems more natural as their will be many keywords.

ankitkalia1195 (approved):

ankitkalia1195:
@malparty Can you check why the Build Docker image (push) is failing, also the unit and system tests can be run again.

malparty:
> @malparty Can you check why the Build Docker image (push) is failing, also the unit and system tests can be run again.

This was because I did not merge from develop before. It's now done, tests are passing ✅ ;-)

config/routes.rb
```diff
@@ -1,3 +1,5 @@
 Rails.application.routes.draw do
+  root to: 'keywords#index'
```
junan, line 2: How about adding a new line after it? So that the `root route` is separated from other routes?
```suggestion
  root to: 'keywords#index'
  
```

junan:
@malparty How about adding the Heroku url in readme? Just like [this](https://github.com/Lahphim/go-crawler-challenge#web-application). It will be easier for other developers to find out the server url without checking from the PR description. 

> Merge branch 'develop' into chore/setup-deployment 

And regarding updating the local branch from the base branch, we use Git's `rebase` instead of `merge` as `rebase` provides cleaner commit history. You can use `rebase` in your nest PRs when you need to get update from the base branch 👍

config/routes.rb
```diff
@@ -1,3 +1,6 @@
...
+
```
junan: Lets remove the trailing whitespace

```suggestion

```

I had a comment also in https://github.com/malparty/google-search-ruby/pull/32#issuecomment-858249183

junan: @malparty I actually meant to just remove the trailing whitespace as the bot commented [here](https://github.com/malparty/google-search-ruby/pull/32#discussion_r648830728)

It was like the bellow screenshot , we just need to remove the trailing whitespace not the new line as we discussed to keep a new line after the `root route` for better readability 
Screen Shot 2021-06-10 at 2 30 15 PM

And after removing the trailing whitespace, it will be like 
Screen Shot 2021-06-10 at 2 23 37 PM

junan (approved):

ankitkalia1195 (approved):

Issue #17, opened 2021-06-07
[API] As a User, I can sign up and sign in/out with a username and password


malparty:
Build API for #5 

## Why?
In order to provide a secure data segregation within the application.

### Additional notes:
Authentication shall be available from web UI as well as API (Oauth 2 RESTful JSON API ).
UI here will be minimal. Advanced UI being done in #16 

## Who Benefits?
Non logged Users

junan:
![Screen Shot 2021-06-07 at 6 10 58 PM](https://user-images.githubusercontent.com/6076762/121010042-17635980-c7bf-11eb-8e1f-ff7672a15722.png)


@malparty Just for your information, Sign up and Logout API is optional for the IC. If you want to skip those you can do. Or if you don't want to skip those, it is also fine 😄 

https://www.notion.so/nimblehq/Internal-Certifications-123ba28275654192ba31011b9e20bed4

PR #35, opened 2021-06-11
[#17] [API] As a User, I can sign up and sign in/out with a username and password

malparty:
[#17 [API] As a User, I can sign up and sign in/out with a username and password](https://github.com/malparty/google-search-ruby/issues/17)

## What happened

Adding OAuth2 API to sign_in, sign_out and sign_up

## Insight

Routes (all POST)
* Sign Up: `/api/v1/users` => `Api::V1::UsersController/create`
* Sign In: `/api/v1/oauth/token` => `Doorkeeper::TokensController/create`
* Sign Out: `/api/v1/oauth/revoke` => `Doorkeeper::TokensController/revoke`
* Manage applications (Doorkeeper UI): `/oauth/applications/` => `Doorkeeper::ApplicationsController/index`

## Proof Of Work

All 3 examples are available in [the postman documentation](https://documenter.getpostman.com/view/16135891/TzeTKqCb) (link available in the GitHub Wiki too)

Following @junan feedback, we now have Doorkeeper UI to manage apps: 

> Note that only the user `id == 1` is authorized. I did not want to add complexity with role managements, it would be to far from the initial need.

image


junan:
@malparty This PR should be against  `feature/user-login-backend` branch instead of `develop` branch.  Because we do first web functionality, then we work on API part based on our web's work.  It also helps reviewers that they dont need to review same same changes across many PRs 👍 


Screen Shot 2021-06-15 at 4 53 18 PM

app/controllers/api/v1/application_controller.rb
```diff
@@ -0,0 +1,18 @@
+# frozen_string_literal: true
+
+module Api
```
ankitkalia1195: Should this be capitalized, as it is acronym. Also it would `API` in rails `ActionController::API`. But for doing that we would need to add this file to config/initializers

```
ActiveSupport::Inflector.inflections do |inflect| 
  inflect.acronym 'API'
end
```

malparty: Very useful to know that!! Thanks!!!

app/controllers/concerns/devise_parameter.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+module DeviseParameter
+  protected
+
+  def update_allowed_parameters
```
junan, line 6: This method has not been used anywhere. May be you need to add following code in the `app/controllers/api/v1/application_controller.rb`

```
 before_action :update_allowed_parameters, if: :devise_controller?
```

app/controllers/concerns/devise_parameter.rb
```diff
@@ -0,0 +1,12 @@
...
+    devise_parameter_sanitizer.permit(:sign_up) { |u| u.permit(:firstname, :lastname, :email, :password) }
```
junan: To match with the database column, it should be like 
```suggestion
    devise_parameter_sanitizer.permit(:sign_up) { |u| u.permit(:first_name, :last_name, :email, :password) }
```

app/controllers/concerns/devise_parameter.rb
```diff
@@ -0,0 +1,12 @@
...
+    devise_parameter_sanitizer.permit(:account_update) do |u|
+      u.permit(:firstname, :lastname, :email, :password, :current_password)
```
junan: Same as above comment

app/controllers/api/v1/application_controller.rb
```diff
@@ -0,0 +1,18 @@
+# frozen_string_literal: true
+
+module API
+  module V1
+    class ApplicationController < ActionController::API
```
junan, line 5: We follow [JSON API](https://jsonapi.org/) standard as per our [convention](https://nimblehq.co/compass/development/code-conventions/#api-design). Based on your postman `users/create`, `oauth/token` API response, its is not aligining  with [JSON API](https://jsonapi.org/) standard.

For example `oauth/token` endpoint response should be like bellow as per  [JSON API](https://jsonapi.org/) standard 
Screen Shot 2021-06-17 at 10 41 10 AM

Also `users/create` endpont should return `201` success status code instead of `200`. For resource creation, it should return `201`.

You can check our Nimble Servey API response on [postman](https://nimblehq.postman.co/workspace/Nimble~9daf6b25-882e-4ad0-8299-95529e2883ff/documentation/11835486-63105f38-c20f-493d-b4d8-873a18c8d8c7)

For referrence, you can check Nimble Servey API codebase([controller](https://github.com/nimblehq/nimble-survey-web/tree/development/app/controllers/api/v1), [concern](https://github.com/nimblehq/nimble-survey-web/tree/development/app/controllers/concerns/api/v1)) which also used Devise, doorkeeper, fast_jsonapi

db/schema.rb
```diff
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2021_06_09_093354) do
+ActiveRecord::Schema.define(version: 2021_06_14_164900) do
```
junan: We have not added any migration in this PR. Why it is changing?

spec/fabricators/application_fabricator.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+Fabricator(:application, from: Doorkeeper::Application) do
+  name 'iOS client'
+  redirect_uri ''
+  scopes ''
```
junan: As we are not setting any default value, I think we can remove it
```suggestion

```

spec/rails_helper.rb
```diff
@@ -22,7 +22,11 @@
 
   config.include Rails.application.routes.url_helpers
 
+  config.include OAuthHelpers
+
   config.include Devise::Test::IntegrationHelpers, type: :system
 
   config.include DeviseHelpers::SystemHelpers
+
+  config.include OAuthHelpers
```
junan: Duplicated, we already added it before

spec/requests/api/v1/users_controller_spec.rb
```diff
@@ -0,0 +1,44 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+describe API::V1::UsersController, type: :request do
+  context 'when a user register' do
+    it 'returns the user' do
+      params = create_user_params
+      params[:email] = 'new_email@gmail.com'
```
junan, line 9:
```suggestion

      params[:email] = 'new_email@gmail.com'
```

spec/requests/api/v1/users_controller_spec.rb
```diff
@@ -0,0 +1,44 @@
...
+      post 'create', params: params
+
+      expect(JSON.parse(response.body)['data'].map { |item| item['type'] }).to contain_exactly('user', 'token')
+    end
+  end
+
+  context 'when a user register with an existing email' do
+    it 'receive an error' do
```
junan:
```suggestion
    it 'receives an error' do
```

spec/requests/api/v1/users_controller_spec.rb
```diff
@@ -0,0 +1,44 @@
...
+      params = create_user_params
+      post 'create', params: params
```
junan:
```suggestion

      post :create, params: params
```

Note: We use Ruby symbol instead of string when possible. As symbol looks better

spec/requests/api/v1/users_controller_spec.rb
```diff
@@ -0,0 +1,44 @@
...
+
+      expect(JSON.parse(response.body).keys).to contain_exactly('error')
+    end
+  end
+
+  context 'when a user register with a non valid password' do
+    it 'receive an error' do
```
junan:
```suggestion
    it 'receives an error' do
```

spec/requests/api/v1/users_controller_spec.rb
```diff
@@ -0,0 +1,44 @@
...
+      params = create_user_params
+      params[:password] = '123'
+      post 'create', params: params
```
junan:
```suggestion

      post :create, params: params
```

spec/requests/api/v1/users_controller_spec.rb
```diff
@@ -0,0 +1,44 @@
...
+
+      expect(JSON.parse(response.body).keys).to contain_exactly('error')
+    end
+  end
+
+  context 'when a user register with a non valid Client Id' do
+    it 'receive an error' do
```
junan:
```suggestion
    it 'receives an error' do
```

spec/requests/api/v1/users_controller_spec.rb
```diff
@@ -0,0 +1,44 @@
...
+      params = create_user_params
+      params[:client_id] = 'not valid'
+      post 'create', params: params
```
junan:
```suggestion

      post :create, params: params
```

spec/requests/doorkeeper/tokens_controller_spec.rb
```diff
@@ -0,0 +1,55 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+describe Doorkeeper::TokensController, type: :request do
```
junan: Similar to above file comments, using symbol, add `s`

app/controllers/concerns/devise_parameter.rb
malparty: We do not use the devise_controller for API (from what I understood). 
Though you are right: there is a problem with this module because the code is duplicated (in the module **and** in the main app_controller)
image

So I will refactor that~

junan: You can use Rails concern feature to fix duplication. Check this `Authorization` module in [Braive](https://github.com/nimblehq/braive-web/blob/development/app/controllers/concerns/authorization.rb)(updated with correct link) project. It is used in the API and Web part

malparty: Ok, it should be fixed! :)

app/controllers/api/v1/users_controller.rb
```diff
@@ -0,0 +1,47 @@
...
+
+        if user.save
+          # create access token for the user, so the user won't need to login again after registration
+          access_token = user.get_access_token client_app.id
+          # return json containing access token and refresh token
+          # so that user won't need to call login API right after registration
+          render json: create_success_result(user, access_token), status: :created
+
+        else
+          render_errors user.jsonapi_errors
```
malparty: I added a method to the model "Users" (via error_handler_concern) in order to parse the model 'errors' into a Json::API compliant error. 
But I'm not sure if this is the right place to be done... (JSON::API is a controller responsibility, not much on the model // but parsing ActiveModel::Error objects seems to be the responsibility of the model... humhum

Any insight @junan  @ankitkalia1195  ? :)
Thanks!

db/schema.rb
malparty: I think this is due to the rebase, maybe I retained a different version number when I did a previous rebase from develop to auth-backend branch. After seen that issue in other rebases, I now try to keep the latest version number when rebasing.

spec/rails_helper.rb
malparty: How couldn't I see it!! (0_o)' 
Thanks!

spec/requests/api/v1/users_controller_spec.rb
malparty: Again I missed ALL the 's' 
thanks

spec/requests/api/v1/users_controller_spec.rb
```diff
@@ -0,0 +1,44 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+describe API::V1::UsersController, type: :request do
+  context 'when a user register' do
```
ankitkalia1195: `when a user registers`

spec/requests/api/v1/users_controller_spec.rb
```diff
@@ -0,0 +1,44 @@
...
+    it 'returns the user' do
+      params = create_user_params
+      params[:email] = 'new_email@gmail.com'
+      post 'create', params: params
+
+      expect(JSON.parse(response.body)['data'].map { |item| item['type'] }).to contain_exactly('user', 'token')
+    end
+  end
+
+  context 'when a user register with an existing email' do
```
ankitkalia1195: `when a user registers with an existing email`

spec/requests/doorkeeper/tokens_controller_spec.rb
```diff
@@ -0,0 +1,55 @@
...
+
+      post 'create', params: params
+
+      expect(JSON.parse(response.body)['error']).to eq('invalid_client')
+    end
+  end
+
+  context 'when user ask for a token with bad credentials' do
+    it 'return invalid_grant error' do
+      params = token_request_params
```
ankitkalia1195: can we simplify it in one line

```
  params = token_request_params.merge(password: 'wrong_pass')
```

spec/requests/api/v1/users_controller_spec.rb
malparty: Thanks! I did some experimentations to better understand "when", it's more clear now :)

spec/requests/doorkeeper/tokens_controller_spec.rb
malparty: the beauty of mastering ruby!! :) 
Nice!! :) 👍

malparty: Applied here too:
image

spec/fabricators/application_fabricator.rb
malparty: working! :)

app/controllers/api/v1/users_controller.rb
junan: @malparty  We usually don't place API related stuff into model. 

I think you can get the related errors with ` user.errors.full_messages.to_sentence`.  Then you can render the error like we did [Survey project](https://github.com/nimblehq/nimble-survey-web/blob/development/app/controllers/api/v1/registrations_controller.rb#L13)
```ruby
  render_error(detail: user.errors.full_messages.to_sentence)
```

malparty: full messages merge all errors into a string, which is a bit sad if we need to assign it to "errors'. 
I'll figure out to have it cleaned :)

app/controllers/api/v1/application_controller.rb
malparty: All API signatures are now JSON API standard.
But Rubocop is raising some warnings (that seems "false positive" IMO):
image

But anyway, it's the time for me to ask: Any other way to be JSON compliant with Doorkeeper::TokenController (without rewritting the whole controller?) :) 

Also I will request access to the postman you sent me: From the source code I cannot believe you have JSON:API valid error messages :P And I do not see where are tokens form handled (maybe I miss a Doorkeeper config?)

malparty: Update: 
- I've put that code in a generic concern and check action names in the filter method itself. 
- Also I have realized that the requests tests were still executed on the Doorkeeper::TokenController (instead of API::V1::TokenController) so I've updated it as well.

But I'm still not understanding 100% how Survey app handle JsonAPI formatting (this app has much cleaner code than my solution so I'd be delighted to know more about it :) )
Thanks!

PS: Talking about 201 code, maybe the team wants to update this action with a 201 status code too :D
image
(as the action "creates a token", it does not log the user technically speaking)

junan: > But I'm still not understanding 100% how Survey app handles JsonAPI formatting 

@malparty Doorkeeper allows to customize token response without overriding everything in the its own token controller.([Customize success token response](https://github.com/doorkeeper-gem/doorkeeper/wiki/Customizing-Token-Response), [Customize error response](https://github.com/doorkeeper-gem/doorkeeper/wiki/Customizing-token-error-response)). In the Nimble Survey PR we did exactly the same thing in this PR you can check it https://github.com/nimblehq/nimble-survey-web/pull/14

> PS: Talking about 201 code, maybe the team wants to update this action with a 201 status code too :D

In the login/token API (`oath/token`), technically it creates a  token in the database, but from the client's perspective(or API perpective)  we are not creating any resource, we just want to login(or just getting a token from the server). Thats why I think Doorkeeper  returns `200` status code for `oath/token` API.  

And to discussion with team and finalizing some conventions, we create a RFC(Request for comment) ticket in the compass, we vote there and update the conventions in the compass. For example, we created [this RFC](https://github.com/nimblehq/compass/issues/642) to finalize Go constant variable convention. You can create a RFC issue for 201/200 status code for `oauth/token` API if you want 👍

malparty: Thanks a lot for awesome clear reply :) 
I will implement the changes now ~ 

For RFC 200/201: well, I believe it's not really important, as you said, the client can either ask "please create me a temporary access_token" or "please give me a way to log in" ^^, 
So I will follow Doorkeeper standard to not add useless config/overwrite ;-)

app/controllers/api/v1/users_controller.rb
malparty: @junan  Okay, code updated, I just created a simple ActiveModel::ErrorsSerializer :) 
So that I can `ErrorsSerializer.new(user.errors).serializable_hash`. 

It seems clean to me~

spec/requests/api/v1/users_controller_spec.rb
```diff
@@ -0,0 +1,44 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+describe API::V1::UsersController, type: :request do
+  context 'when a user registers' do
+    it 'returns the user' do
+      params = create_user_params
+      params[:email] = 'new_email@gmail.com'
+      post :create, params: params
```
ankitkalia1195: can we simplify to one line

```
   post :create, params: create_user_params.merge!(email: '')
```

spec/requests/api/v1/users_controller_spec.rb
```diff
@@ -0,0 +1,44 @@
...
+
+      expect(JSON.parse(response.body).keys).to contain_exactly('errors')
+    end
+  end
+
+  context 'when a user registers with a non valid password' do
+    it 'receives an error' do
+      params = create_user_params
+      params[:password] = '123'
```
ankitkalia1195: Same comment as above

spec/requests/api/v1/users_controller_spec.rb
```diff
@@ -0,0 +1,44 @@
...
+      post :create, params: params
+
+      expect(JSON.parse(response.body).keys).to contain_exactly('errors')
+    end
+  end
+
+  context 'when a user registers with a non valid Client Id' do
+    it 'receives an error' do
+      params = create_user_params
+      params[:client_id] = 'not valid'
```
ankitkalia1195: Same comment as above

spec/support/oauth_helpers.rb
```diff
@@ -0,0 +1,55 @@
+# frozen_string_literal: true
+
+module OAuthHelpers
+  def token_request_params
+    @user ||= Fabricate(:user)
```
ankitkalia1195, line 5: @malparty I guess we can use fabricators instead of creating params for these. For fabricators, we can also create them without a class. Refer to this example in the1-api. 

https://github.com/nimblehq/the1-api/blob/development/spec/fabricators/v2/token.rb

ankitkalia1195: This one is not extremely important, but according to your preference :)

app/serializers/doorkeeper/token_serializer.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+module Doorkeeper
+  class TokenSerializer
+    include JSONAPI::Serializer
+    attributes :token, :token_type, :expires_in, :refresh_token
```
ankitkalia1195, line 7: we can have a line between `attributes` and `attribute`, as we provide a block to it and remaining are just single line

spec/support/oauth_helpers.rb
malparty: When fabricating a token, can it be valid? 
From my tests, revoking a fabricated tokens result in an error as that stub is not valid. 

So I'm not sure where exactly we could use a stub:
- test: 'ask for a token' = we are actually testing to query a token
- test 'when a user refreshes an existing token on time' = we need a valid token
- test 'when a user revokes an existing token' = same

Or maybe I did not understand something ~

app/serializers/doorkeeper/token_serializer.rb
malparty: Done ;-)

app/controllers/api/v1/tokens_controller.rb
```diff
@@ -0,0 +1,29 @@
...
+
+      # Overwrite this method as it bypass the custom_error_response provided by Doorkeeper
+      def revocation_error_response
+        error_description = I18n.t(:unauthorized, scope: %i[doorkeeper errors messages revoke])
+        {
+          errors: build_error(detail: error_description, code: :invalid_client)
+        }
+      end
+
+      def handle_jsonapi_success
```
ankitkalia1195: @malparty Firstly, if you need a callback only specific to a action, you can 
```
   after_action : handle_jsonapi_success, only: :revoke
```

Secondly, IMO going this way  of setting response in after_action is a bit confusing and we usually dont do this in rails proejcts. If they dont provide a single method we could override, we can override the whole action `revoke`.

app/controllers/api/v1/users_controller.rb
```diff
@@ -0,0 +1,44 @@
...
+module API
+  module V1
+    class UsersController < API::V1::ApplicationController
+      include API::V1::ErrorHandlerConcern
+
+      skip_before_action :doorkeeper_authorize!, only: :create
+
+      def create
+        user = User.new(create_params.except(:client_id))
+        client_app = Doorkeeper::Application.find_by(uid: create_params[:client_id])
```
ankitkalia1195: can this client app part be moved to a before callback, then you wont need to use return as well. Also check if doorkeeper provides some method for it.

```
   before_action :ensure_valid_client, only: :create
 
   def ensure_valid_client
     @client_app = Doorkeeper::Application.find_by(uid: create_params[:client_id])

     render_error 'Invalid client ID', status: :forbidden, source: :client_id if @client_app.blank?
   end
end

File: app/controllers/api/v1/tokens_controller.rb (hunk 26 above)
Comment (malparty) [reply app/controllers/api/v1/tokens_controller.rb]: Yes, I had a first version with the `only: : revoke` keyword. But Rubocop does not allow it (as the revoke action is not defined in the same file = Rubocop indicates that this is confusing for developers). Which I understand. 

Point 2: Yes, I agree it's an ugly hack!! And so I'll go for the re-write of the whole action :) Thanks!

Comment (malparty) [reply app/controllers/api/v1/tokens_controller.rb]: Ok, done @ankitkalia1195  ;-)

File: app/models/user.rb [hunk 28]
```diff
@@ -11,4 +11,25 @@ def self.authenticate(email, password)
     user = User.find_for_authentication(email: email)
     user&.valid_password?(password) ? user : nil
   end
+
+  def get_access_token(client_app_id)
```
ankitkalia1195: The method name can be `create_access_token`, as we are creating new everytime we call this.

app/controllers/api/v1/users_controller.rb
malparty: Yes we can @ankitkalia1195  (as we only need the client id inside the create action, so we can get it from params again).
But is it really worth it? As we do not expect to have more actions in this controller ; isn't it trying to solve a problem that is not here? (I'm asking as I probably don't know about something, so I want to be sure to understand the why -- in order to not reproduce the same mistake twice)
Thanks!! :)

app/controllers/api/v1/users_controller.rb
```diff
@@ -0,0 +1,44 @@
...
+        # return json containing access token and refresh token
+        # so that user won't need to call login API right after registration
+        render json: create_success_result(user, access_token), status: :created
+      end
+
+      private
+
+      def create_success_result(user, access_token)
+        user_json = UserSerializer.new(user).serializable_hash
+        token_json = Doorkeeper::TokenSerializer.new(access_token).serializable_hash
```
ankitkalia1195: @malparty How about remove this method entirely by creating a new serializer, by having token as a part of user.

```
    UserTokenSerializer < UserSerializer
       attribute :access_token

        def access_token
              @access_token ||= object.create_access_token
       end
end
    
Then in our controller, we can simply

```
    render json: user, serializer: UserTokenSerializer
```

File: app/models/user.rb (hunk 28 above)
Comment (malparty) [reply app/models/user.rb]: Make perfect sense! 📝
Fixed ;-)

File: app/controllers/api/v1/users_controller.rb (hunk 27 above)
Comment (ankitkalia1195) [reply app/controllers/api/v1/users_controller.rb]: @malparty Firstly, our action create would become shorter and look much cleaner. Secondly, this is something which can be reusable across other actions in future. For instance, have a look at doorkeeper tokens controller https://github.com/doorkeeper-gem/doorkeeper/blob/main/app/controllers/doorkeeper/tokens_controller.rb. In that, they do something similar, even though there is only one action.

```
    before_action :validate_presence_of_client, only: [:revoke]
```

File: app/controllers/api/v1/users_controller.rb (hunk 29 above)
Comment (malparty) [reply app/controllers/api/v1/users_controller.rb]: It would change the result signature (token being nested inside the user object -- is it okay for JsonApi? If I get it right it only prevent us to specify an explicit relationship between the user and his token), but I like the "cleanness" of this idea!!! :) 

I will give it a try before monday!! :) // Need to go now =/

File: app/controllers/api/v1/tokens_controller.rb [hunk 30]
```diff
@@ -0,0 +1,40 @@
+# frozen_string_literal: true
+
+module API
+  module V1
+    class TokensController < Doorkeeper::TokensController
+      include ErrorHandlerConcern
+
+      def revoke
```
junan, line 9: Why you need to override `revoke` action?  🤔  Usually doorkeeper return empty response which if the revoke success. Or you want to show some JSON API response?

app/controllers/api/v1/users_controller.rb
```diff
@@ -0,0 +1,44 @@
...
+          return
+        end
+
+        render_errors ActiveModel::ErrorsSerializer.new(user.errors).serializable_hash and return unless user.save
+
+        # create access token for the user, so the user won't need to login again after registration
+        access_token = user.create_access_token client_app.id
+        # return json containing access token and refresh token
+        # so that user won't need to call login API right after registration
+        render json: create_success_result(user, access_token), status: :created
```
junan: I think we need to use user serialized here similar to Ankit comment [here](https://github.com/malparty/google-search-ruby/pull/35/files#r654332811)
```ruby
  # this is pseudo code
  render json: UserSerializer.new(user)
  Or 
  # if you want to just response user token information(Based on Ankit comment)
  render json: UserTokenSerializer.new(user)
```
We mainly render json response using serializer. Like we did in our previous projects([Servey](https://github.com/nimblehq/nimble-survey-web/blob/development/app/controllers/api/v1/surveys_controller.rb#L11), [The-1](https://github.com/nimblehq/the1-api/blob/development/engines/the1_idm/app/controllers/the1_idm/v2/accounts_controller.rb#L18))

app/serializers/doorkeeper/token_serializer.rb
```diff
@@ -0,0 +1,12 @@
...
```
junan, line 7:
```suggestion

    attributes :token, :token_type, :expires_in, :refresh_token
```

app/serializers/user_serializer.rb
```diff
@@ -0,0 +1,6 @@
+# frozen_string_literal: true
+
+class UserSerializer
+  include JSONAPI::Serializer
```
junan, line 4:
```suggestion
  include JSONAPI::Serializer
  
```

spec/support/oauth_helpers.rb
```diff
@@ -0,0 +1,55 @@
...
+      client_secret: @application.secret,
+      email: @user.email,
+      password: @user.password
+    }
+  end
+
+  def query_token
+    params = token_request_params
+    post 'create', params: params
+    JSON.parse(response.body)
```
junan:
```suggestion
    params = token_request_params
    
    post :create, params: params
    
    JSON.parse(response.body)
```

spec/support/oauth_helpers.rb
```diff
@@ -0,0 +1,55 @@
...
+      email: @user.email,
+      password: @user.password,
+      refresh_token: refresh_token
+    }
+  end
+
+  def query_refresh_token(refresh_token)
+    params = token_refresh_params refresh_token
+    post 'create', params: params
+    JSON.parse(response.body)
```
junan:
```suggestion
    params = token_refresh_params refresh_token
    
    post :create, params: params
    
    JSON.parse(response.body)
```

spec/support/oauth_helpers.rb
```diff
@@ -0,0 +1,55 @@
...
+
+  def create_user_params
+    @user ||= Fabricate(:user)
+    @application ||= Fabricate(:application)
+
+    {
+      password: @user.password,
+      email: @user.email,
+      lastname: @user.last_name,
+      firstname: @user.first_name,
```
junan:
```suggestion
      last_name: @user.last_name,
      first_name: @user.first_name,
```

app/controllers/api/v1/tokens_controller.rb
malparty: Yes @junan, empty reponse is not compliant with JSON API standard :P That's the cleanest way we found with Ankit ^^

spec/support/oauth_helpers.rb
malparty: Nice catch!! :O Thanks! :)

app/controllers/api/v1/users_controller.rb
malparty: okay, I've submited a proposal. 
It's not as clean as your solution because we need to pass the ClientId in order to create the new token on the fly. 
A simplest solution used in most of the other IC is to NOT send back a token when the user just signed in. But as the purpose is to learn, I think it's great to dive into this :)

malparty: @ankitkalia1195   @junan  feel free to comments in here 
(Junan, I will resolve the other comment quite related to this thread)

app/controllers/api/v1/users_controller.rb
malparty: yes, cf other thread ;-)

app/controllers/api/v1/users_controller.rb
malparty: Thanks a lot! I could not see it clearly before working on the serializer, but not the "before/after" of the create method is impressive! Much clearer! :)

spec/support/oauth_helpers.rb
ankitkalia1195: @malparty In this my idea was this, you do not fabricate the token, but only the params. Then in test you can get params from this fabricator. The main advantage here is that we will not need to use ruby merge, and can use fabricator syntax to override any param. But as i said, its optional and does not provide a big benefit 👍 

```
    Fabricate(:token_params, from: 'OpenStruct') do
       
    end
```

```
    params = Fabricate(:token_params).to_h
```

spec/support/oauth_helpers.rb
```diff
@@ -0,0 +1,59 @@
...
+
+  def create_user_params
+    @user ||= Fabricate(:user)
+    @application ||= Fabricate(:application)
+
+    {
+      password: @user.password,
+      email: @user.email,
+      last_name: @user.last_name,
+      firstname: @user.first_name,
```
ankitkalia1195: Why is this without underscore? `_`

app/controllers/api/v1/users_controller.rb
```diff
@@ -0,0 +1,34 @@
...
+      end
+
+      private
+
+      def create_params
+        params.permit(:email, :password, :last_name, :first_name, :client_id)
+      end
+
+      def ensure_valid_client
+        @client_app = Doorkeeper::Application.find_by(uid: create_params[:client_id])
```
junan: I am not sure only validating `client_id` is the best practice or not. Its seems it is better to validate both `client_id`, `secret` for better security .
```suggestion
        @client_app = Doorkeeper::Application.by_uid_and_secret(params[:client_id], params[:client_secret])
```

In the Survey api, it required both for registration also. Wjat do you think?

app/controllers/api/v1/application_controller.rb
```diff
@@ -0,0 +1,19 @@
...
```
junan, line 5: Top access API, the client first need to access `client_id`, `secret`.  How s/he will get it?

In our Go certification, we mostly [created](https://github.com/Lahphim/go-crawler-challenge/pull/69) a view, where logged in user can generate the `client_id` and `secret`.  For Doorkeeper, it does everything. You might need to change the route setting or so. What do you think?

spec/support/oauth_helpers.rb
```diff
@@ -0,0 +1,59 @@
...
+      client_secret: @application.secret,
+      email: @user.email,
+      password: @user.password
+    }
+  end
+
+  def query_token
+    params = token_request_params
+
+    post 'create', params: params
```
junan:
```suggestion
    post :create, params: params
```

spec/support/oauth_helpers.rb
```diff
@@ -0,0 +1,59 @@
...
+      email: @user.email,
+      password: @user.password,
+      refresh_token: refresh_token
+    }
+  end
+
+  def query_refresh_token(refresh_token)
+    params = token_refresh_params refresh_token
+
+    post 'create', params: params
```
junan:
```suggestion
    post :create, params: params
```

spec/requests/api/v1/users_controller_spec.rb
```diff
@@ -0,0 +1,37 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+describe API::V1::UsersController, type: :request do
+  context 'when a user registers' do
+    it 'returns the user' do
+      post :create, params: create_user_params.merge!({ email: 'new_email@gmail.com' })
```
junan: We can omit `!` and `{}`
```suggestion
      post :create, params: create_user_params.merge(email: 'new_email@gmail.com' )
```

spec/requests/api/v1/users_controller_spec.rb
```diff
@@ -0,0 +1,37 @@
...
+    it 'receives an error' do
+      post :create, params: create_user_params
+
+      expect(JSON.parse(response.body).keys).to contain_exactly('errors')
+    end
+  end
+
+  context 'when a user registers with a non valid password' do
+    it 'receives an error' do
+      post :create, params: create_user_params.merge!({ password: '123' })
```
junan: Same as above comment

spec/requests/api/v1/users_controller_spec.rb
```diff
@@ -0,0 +1,37 @@
...
+
+      expect(JSON.parse(response.body).keys).to contain_exactly('errors')
+    end
+  end
+
+  context 'when a user registers with a non valid Client Id' do
+    it 'receives an error' do
+      post :create, params: create_user_params.merge!({ client_id: 'not valid' })
```
junan: Same as above comment

app/controllers/api/v1/application_controller.rb
malparty: I think it depends on the specifications (or say, the client needs) :) 
Here, the IC scenario says that WE are going to build a mobile app out of that, so my assumption is that the Client Id does not need public endpoints. 

But let's be player, I will update the code as if we are in a situation where other people/organizations might want to consume ou API. ;-)

app/controllers/api/v1/users_controller.rb
malparty: Makes good sense~ Fixed ;-)

spec/requests/api/v1/users_controller_spec.rb
malparty: I will remove the parentheses, though, for the `!` I had the same reaction as you when reading @ankitkalia1195  recommendation, but after some thought, I felt like "if `merge!` does not create a second variable, so it's probably more performant". 
As you raised the concern again, I decided to search for it :) 
So thanks to you, I've learned :) 
https://www.paperlesspost.com/blog/teams/different-methods-for-merging-ruby-hashes/

So even for super-small operations, I believe it's good to prefere merge! when it does not change anything else in the code :)

malparty: Updated & pushed ;-)

spec/support/oauth_helpers.rb
malparty: Oh, okay now I understand the idea!! 
I will try to apply it for the next features if it's ok for you!! :) 
--> I did not know we can override params with the Fabricate(..) method directly, thanks!!!

app/controllers/api/v1/application_controller.rb
malparty: All right, interesting adventure :) It's done & I have also research on how to manually create/edit applications using Doorkeeper library (might be useful for future projects). 

The description of this change has been added in the PR description ;-)

Thanks!

spec/support/oauth_helpers.rb
malparty: Rebase hell. I've fixed it among some other bad side effects.

app/controllers/api/v1/users_controller.rb
```diff
@@ -0,0 +1,35 @@
...
+      skip_before_action :doorkeeper_authorize!, only: :create
+
+      before_action :ensure_valid_client, only: :create
+
+      def create
+        user = User.new(create_params.except(:client_id, :client_secret))
+
+        render_errors ActiveModel::ErrorsSerializer.new(user.errors).serializable_hash and return unless user.save
+
+        render json: UserTokenSerializer.new(user, { params: { client_id: @client_app.id } }).serializable_hash,
```
ankitkalia1195: @malparty I guess it should work without calling `serializable_hash` as well? Can you try it here?

config/initializers/doorkeeper.rb
```diff
@@ -58,6 +60,15 @@
   #     redirect_to sign_in_url
   #   end
   # end
+  #
+  # WARNING, the following code allow any user to edit Client Application endpoints!
+  admin_authenticator do
+    if current_user
+      head :forbidden unless current_user.id == 1
```
ankitkalia1195: I guess adding boolean to the user should be better right? Comparing `id == 1` feels so weird. Altough this security practice of restricting user access to doorkeeper panel is not required in the IC, but if we add it, the implementation should be ok. So we can either remove it or add it the proper way :)

malparty: "if we add it, the implementation should be ok" ==> Well said! I like that! :) 

Done ✅

app/controllers/api/v1/users_controller.rb
malparty: Tried and YES, it works inside as render section. Though it does not work in other places (like UserTokenSerializer). 

Good to know! :) 

And... fixed! ✅

config/initializers/doorkeeper.rb
```diff
@@ -58,6 +60,15 @@
   #     redirect_to sign_in_url
   #   end
   # end
+  #
+  # WARNING, the following code allow any user to edit Client Application endpoints!
+  admin_authenticator do
+    if current_user
```
ankitkalia1195, line 66: Can we simplify it to `current_user&.is_admin?`. Also we can add predicate `?` to method when using boolean values as per our conventions .(Rails add predicate method by default for boolean values).

db/migrate/20210621104559_add_is_admin_to_users.rb
```diff
@@ -0,0 +1,5 @@
+class AddIsAdminToUsers < ActiveRecord::Migration[6.1]
+  def change
+    add_column :users, :is_admin, :boolean, default: false
```
ankitkalia1195: How about adding `null: false` as well :).

ankitkalia1195: We can add index to it as well as per our conventions https://nimblehq.co/compass/development/code-conventions/ruby-on-rails/#relational-database

config/initializers/doorkeeper.rb
malparty: Good point for the `?` predicate @ankitkalia1195  :) 

Regarding `current_user&.is_admin?`, I think it's better to keep the 2 different paths:
1. If user is not logged = redirect to sign_in
2. If user is logged, then we check if he shall gets a 403 or not

As an unauthenticated admin user, it's more user-friendly :)

app/serializers/doorkeeper/token_serializer.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+module Doorkeeper
+  class TokenSerializer
+    include JSONAPI::Serializer
+
+    attributes :token, :token_type, :expires_in, :refresh_token
+
+    attribute :created_at do |token|
+      token.created_at.to_time.to_i
```
ankitkalia1195: How about directly calling `.to_i`, why do we need to convert to time

```
   token.created_at.to_i
```

malparty: Copy/paste from doorkeeper tutorial >< 
I just removed it :)

db/migrate/20210621104559_add_is_admin_to_users.rb
malparty: True!! 
Updated, thanks!!!

Gemfile
```diff
@@ -53,7 +54,6 @@ group :development, :test do
   gem 'rubocop-rails', require: false # A RuboCop extension focused on enforcing Rails best practices and coding conventions.
   gem 'rubocop-rspec', require: false # Code style checking for RSpec files
   gem 'rubocop-performance', require: false # An extension of RuboCop focused on code performance checks.
-  gem 'ffaker' # used to easily generate fake data: names, addresses, phone numbers, etc.
```
junan, line 56: Why we are removing it? 🤔  it easy to generate test data with it

config/initializers/doorkeeper.rb
ankitkalia1195: Sure, makes sense :)

app/controllers/api/v1/users_controller.rb
```diff
@@ -0,0 +1,35 @@
+# frozen_string_literal: true
+
+module API
+  module V1
+    class UsersController < API::V1::ApplicationController
```
junan: I think it should work without the `API::V1::` namespace
```suggestion
    class UsersController < ApplicationController
```

app/controllers/api/v1/application_controller.rb
```diff
@@ -0,0 +1,19 @@
...
+      include DeviseParameter
```
junan: I think this one is not needed for API, even `update_allowed_parameters`  is not used in the in any API. What do you think?

app/models/user.rb
```diff
@@ -11,4 +11,25 @@ def self.authenticate(email, password)
     user = User.find_for_authentication(email: email)
     user&.valid_password?(password) ? user : nil
   end
+
+  def create_access_token(client_app_id)
```
junan, line 15: Does it works? I think `@id` value will be `nil` here

spec/rails_helper.rb
junan: We already added it 😄

spec/requests/api/v1/users_controller_spec.rb
```diff
@@ -0,0 +1,45 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+describe API::V1::UsersController, type: :request do
+  context 'when a user registers' do
+    it 'returns the user' do
+      post :create, params: create_user_params.merge!(email: 'new_email@gmail.com')
```
junan: We dont need `!`, please check bellow test cases also 👍 
```suggestion
      post :create, params: create_user_params.merge(email: 'new_email@gmail.com')
```

config/initializers/doorkeeper.rb
malparty: Ok. predicate added. 
I keep this conversation unresolve so that @ankitkalia1195 you can confirm if my previous comment is ok for you, or not :) Thanks

Gemfile
malparty: Because duplicated... It is still here [earlier in the GemFile](https://github.com/malparty/google-search-ruby/blob/f4fd89d668c0894e99a8a5323aaedfd21f4cd020/Gemfile#L11)

spec/rails_helper.rb
malparty: I hate rebase... >< 
And trust that I've been mega careful in every conflict merging... I'm wondering if it simply merged it without seen the conflict (I probably added this line in 2 earlier branches).

app/models/user.rb
malparty: :O 
Yes, totally true! Fixed!

app/controllers/api/v1/application_controller.rb
malparty: It makes no sense indeed!! I'm wondering how I could let such a thing here!
Thanks a lot!! :)

app/controllers/api/v1/users_controller.rb
malparty: Yes, true!
Tested here:
image

Thanks! :)

app/models/user.rb
```diff
@@ -14,7 +14,7 @@ def self.authenticate(email, password)
 
   def create_access_token(client_app_id)
     Doorkeeper::AccessToken.create(
-      resource_owner_id: @id,
+      resource_owner_id: self[:id],
```
ankitkalia1195: @malparty Just `id` or `self.id` can also work :)

config/initializers/doorkeeper.rb
malparty: Meanwhile I saw your comment "sure makes sense", so I will mark the conversation as resolve :) Thanks! 🙏

app/models/user.rb
malparty: >< I forgot to reload the rails console when testing this option so it failed. But now trying again, yes, much easier! :)

spec/requests/api/v1/users_controller_spec.rb
malparty: @ankitkalia1195 interesting feedback from Junan, what do you think? :) 

I do not have any problem by updating to the no-bang method (as discussed with Junan, performances won't change much in a single-call and so-small hash like this), but I'm not convinced 100% by the risk of this `merge!` in this specific context. 
It's also to better understand the general idea of when does it makes sense to risk the bang-method and when we should prefer the non-bang ones :) 
Thanks

app/serializers/active_model/errors_serializer.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+module ActiveModel
+  class ErrorsSerializer
+    def initialize(errors)
+      @errors = errors
+    end
+
+    def serializable_hash
```
ankitkalia1195: Maybe we can try to rename `serializeable_hash` to `as_json`, then controller call would be simple.

```
  def as_json
     { errors: @errors.errors.map { |e| { detail: e.full_message, source: e.attribute } } }
  end
```

```
   render json: ActiveModel:: ErrorsSerializer.new(user.errors)
```

Above change is easy, if you want to go step further (optional), you can even override ActiveModel::Errors class https://apidock.com/rails/ActiveModel/Errors/as_json. Then the controller call will be as simple as:

```
   render json: user.errors
```

malparty: 🔥🤯🔥
My head when I finaly understood that render json:  will call the "as_json" method 🙃

I'll go for the first [simpler] option you mentioned :)

malparty: Done. Any comment is welcome @ankitkalia1195  :)

app/controllers/api/v1/application_controller.rb
```diff
@@ -0,0 +1,17 @@
...
+module API
+  module V1
+    class ApplicationController < ActionController::API
+      # equivalent of authenticate_user! on devise, but this one will check the oauth token
+      before_action :doorkeeper_authorize!
+
+      private
+
+      # helper method to access the current user from the token
+      def current_user
```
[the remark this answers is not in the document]
malparty: Will be needed later ;-)

spec/requests/api/v1/users_controller_spec.rb
```diff
@@ -0,0 +1,45 @@
...
+
+  context 'when a user registers with a non valid password' do
+    it 'receives an error' do
+      post :create, params: create_user_params.merge!(password: '123')
+
+      expect(JSON.parse(response.body).keys).to contain_exactly('errors')
+    end
+  end
+
+  context 'when a user registers with a non valid Client Id' do
```
ankitkalia1195: because context 'when a user registers' is already a parent, can we just use shorter form `given an invalid client id` :)

malparty: @ankitkalia1195  You mean I can have nested contexts? Like that:
``` ruby
context 'when a user registers' do
    context 'given an invalid client id' do
        it '...' do
           [...]
        end
    end
end
```

app/controllers/api/v1/users_controller.rb
```diff
@@ -0,0 +1,35 @@
...
+module API
+  module V1
+    class UsersController < ApplicationController
+      include API::V1::ErrorHandlerConcern
+
+      skip_before_action :doorkeeper_authorize!, only: :create
+
+      before_action :ensure_valid_client, only: :create
+
+      def create
```
junan, line 12: We usually [write](https://github.com/nimblehq/braive-web/blob/development/app/controllers/api/v1/journals_controller.rb#L19) simple `if/else` for rendering success/fail response instead of early return for the controller which makes more readable. How about wrap it under `if/else` block?
```Ruby
        user = User.new(create_params.except(:client_id, :client_secret))

        if user.save
          render json: UserTokenSerializer.new(user, { params: { client_id: @client_app.id } }),
                 status: :created
        else
          render json: ActiveModel::ErrorsSerializer.new(user.errors)
        end
```

app/controllers/api/v1/tokens_controller.rb
ankitkalia1195, line 9: Maybe we can add a comment at top of this function

```
   # Overriden from doorkeeper as the doorkeeper revoke action does not return response according to json-api spec
```

db/seeds.rb
```diff
@@ -12,5 +12,3 @@
   Doorkeeper::Application.create(name: "iOS client", redirect_uri: "", scopes: "")
   Doorkeeper::Application.create(name: "Android client", redirect_uri: "", scopes: "")
 end
-
-Fabricate(:user)
```
junan, line 16: Is it rebase conflict issue?

spec/requests/api/v1/users_controller_spec.rb
```diff
@@ -0,0 +1,45 @@
...
+    it 'receives an error' do
+      post :create, params: create_user_params.merge!(client_id: 'not valid')
+
+      expect(JSON.parse(response.body).keys).to contain_exactly('errors')
+    end
+  end
+
+  context 'when a user registers with a non valid Client Secret' do
```
ankitkalia1195: Same comment as above

ankitkalia1195:
@malparty Last few minor comments, very close to approving it. :)

app/controllers/api/v1/users_controller.rb
ankitkalia1195: @junan @malparty Yes indeed and it looks cleaner for controller as well, I would prefer going this way as well. Altough @malparty might have been confused with this convention on compass `Prefer using guard clauses over if/else and ternary operators.`, but for controller actions we mostly go for if/else.

spec/requests/api/v1/users_controller_spec.rb
ankitkalia1195: @malparty Yes we can have nested contexts.

app/controllers/api/v1/users_controller.rb
ankitkalia1195: @malparty I know it might be confusing, but to summarize:

```
    * Prefer using if/else for situation like this (rendering inside controller action)
    * Prefer guard clause for most other scenarios
 ```

Comment (ankitkalia1195) [reply app/controllers/api/v1/users_controller.rb:12]: Also notice that the convention on compass is prefer, it does not say avoid. We can use `if/else` and `guard` clauses based on the readability of code.

File: app/controllers/api/v1/tokens_controller.rb (hunk 30 above)
Comment (malparty) [reply app/controllers/api/v1/tokens_controller.rb:9]: Also updated comments in the other method overridden ;-)

File: spec/requests/api/v1/users_controller_spec.rb [hunk 62]
```diff
@@ -0,0 +1,47 @@
...
+
+    context 'given a non valid client_id' do
+      it 'receives an error' do
+        post :create, params: create_user_params.merge!(client_id: 'not valid')
+
+        expect(JSON.parse(response.body).keys).to contain_exactly('errors')
+      end
+    end
+
+    context 'given a non valid client_secret' do
```
ankitkalia1195: `given an invalid client_secret`

spec/requests/api/v1/users_controller_spec.rb
```diff
@@ -0,0 +1,47 @@
...
+
+    context 'given a non valid password' do
+      it 'receives an error' do
+        post :create, params: create_user_params.merge!(password: '123')
+
+        expect(JSON.parse(response.body).keys).to contain_exactly('errors')
+      end
+    end
+
+    context 'given a non valid client_id' do
```
ankitkalia1195: `given an invalid client_id`

spec/requests/api/v1/users_controller_spec.rb
```diff
@@ -0,0 +1,47 @@
...
+
+    context 'given an existing email' do
+      it 'receives an error' do
+        post :create, params: create_user_params
+
+        expect(JSON.parse(response.body).keys).to contain_exactly('errors')
+      end
+    end
+
+    context 'given a non valid password' do
```
ankitkalia1195: `given an invalid password`

spec/requests/api/v1/users_controller_spec.rb
malparty: Thanks! :) 
Done for authentication specs. 
I will also review the other specs in other branches.

app/controllers/api/v1/users_controller.rb
malparty: In fact my confusion came from Rubocop who was not happy with the traditionnal if/else. But this was due to my initial code style -- now with this approach Rubocop allows me a normal if/else. 

It's fixed! ;-) And yeah, clearer ~ 

Thanks!!!

db/seeds.rb
malparty: @junan While testing again with new db, I realize this is not needed (now that I better understand how tests works). So I removed it inside this branch as it's still related to authentication.

spec/requests/api/v1/users_controller_spec.rb
malparty: Solved everywhere :)

spec/requests/api/v1/users_controller_spec.rb
ankitkalia1195: But I guess for this comment from junan makes sense, by default it should be merge(less risk) and when you require to make change to the same hash, then it should be merge!. Altough in your case there is no risk, but making a habit to use ! can be risky in this situation:

```
  params = user_params

  post :some_action, params: params.merge!(a: 'b')

  post :another_action, params: params.merge!(b: 'c')

```

Or suppose you are passing hash as argument to some function, where you unintentionally modify hash without :

```
  def do_something(params)
    params.merge!(a: 'b')
  end

  do_something(params)

```

So to summarize:

```
  * Use bang method when you actually need to modify the same hash object in place
  * Use non-bang method otherwise

```

malparty: Ok, it's similar to Junan feedbacks in another discussion~ So I need to change my habits :P 
Thanks guys! And sorry Junan for having been a little bit sceptical, it change a lot from what I used to do~

malparty: Ok solved ;-)

spec/requests/api/v1/users_controller_spec.rb
malparty: All reviewed :)

ankitkalia1195 (approved):

junan (approved):
LGTM 🎉

malparty:
> LGTM 🎉

@Junan, "Looks Good To Me" or "Let's Go To Merge"? 😄

spec/requests/api/v1/users_controller_spec.rb
junan: @malparty Its fine, actually it is good practice to always verify the changes rather than just following feedback 👍

PR #36, opened 2021-06-14
#6 As a user, I can query a single keyword and get its Google search raw response

### LINKED ISSUE #6: As a user, I can query a single keyword and get its Google search raw response
## Why? 
In order for the user to **manually** analyse results (AdWords, links, etc...).

### Additional notes:
Problems to solve:
- AdWords link are included in the result (google remove adwords for bot requests!!)
- Each query executes in less than 1s
- Results come from https://google.com/search 

No Gos
- No data persistence [yet]
- No query parsing [yet]
- No CSV File upload/parsing [yet]
- No fancy UI as such UI would not be sustained

## Who Benefits?
Logged users.

malparty:
[#6 As a user, I can query a single keyword and get its Google search raw response](https://github.com/malparty/google-search-ruby/issues/6)

## What happened

GoogleService::ClientService added to perform simple Google query.

## Insight

- Keywords#create added so that the user can query any keyword from the home page
- Create view created to show the raw result (as nothing is parsed for this task)
- Tests use a VCR cassette that reproduce a query. Note the usage of "vpn" keyword for the cassette query, this ensure to have AdWords in the result

To test: Run the app, perform a search and enjoy the google result.

## Proof Of Work

| Search form added: | Result include Ads: |
|---|---|
| image | image |


spec/services/google_service/client_service_spec.rb
```diff
@@ -0,0 +1,25 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe GoogleService::ClientService, type: :service do
+  context 'when querying a simple keyword' do
+    it 'returns an HTTParty Response' do
```
ankitkalia1195: Can we add the cassete in the test itself
```
it 'returns an HTTParty Response', vcr: 'google_search' do
end
```

app/controllers/keywords_controller.rb
```diff
@@ -2,4 +2,9 @@
 
 class KeywordsController < ApplicationController
   def index; end
+
+  def create
+    @keyword = params['keyword']
+    @raw_response = GoogleService::ClientService.query(@keyword)
```
ankitkalia1195: In case of failure, I guess we should redirect to index with error.

app/views/keywords/create.html.erb
```diff
@@ -0,0 +1,4 @@
+<h1>Result for <%= @keyword %></h1>
```
ankitkalia1195: @malparty According to our compass conventions, we should not use instance variables in views. Instead pass them as local variables from controllers:

```
   render :create, keyword: @keyword, raw_response: @raw_response
```

spec/services/google_service/client_service_spec.rb
malparty: Much clearer tests! Thanks!

app/views/keywords/create.html.erb
malparty: Oh yeah! Thanks!

app/controllers/keywords_controller.rb
malparty: OKay, I gave it a try! (pushed in this branch already). 
Any feedback is of course welcome :)

app/controllers/keywords_controller.rb
```diff
@@ -2,4 +2,19 @@
 
 class KeywordsController < ApplicationController
   def index; end
+
+  def create
```
ankitkalia1195: @malparty As per our new discussion, do we need this action. Will it be a feature used by the end user of the application? https://github.com/malparty/google-search-ruby/issues/6

app/controllers/keywords_controller.rb
malparty: @ankitkalia1195  ready for re-review ;-)

app/controllers/keywords_controller.rb
malparty: @ankitkalia1195  Nope, I used it a lot to preview the results and later to build the parser, but after getting used to vcr cassette and using pre-recorded html content, it's indeed useless. 
I've just removed it (as well as the form in the index view).

app/services/google_service/client_service.rb
```diff
@@ -0,0 +1,41 @@
+# frozen_string_literal: true
+
+module GoogleService
+  class ClientService
+    require 'httparty'
+    require 'colorize'
```
junan: It should load automatically 👍 
```suggestion

```

app/services/google_service/client_service.rb
```diff
@@ -0,0 +1,41 @@
...
+
+    USER_AGENT = 'Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) '\
+                'AppleWebKit/537.36 (KHTML, like Gecko) Chrome/91.0.4472.77 Safari/537.36'
+
+    def initialize(keyword, lang = 'en')
+      escaped_keyword = CGI.escape(keyword)
+      @uri = URI("https://google.com/search?q=#{escaped_keyword}&hl=#{lang}&gl=#{lang}")
+    end
+
+    def query_result
```
junan: Based on out convention, this method name should be `call` https://nimblehq.co/compass/development/code-conventions/ruby-on-rails/#services

app/services/google_service/client_service.rb
```diff
@@ -0,0 +1,41 @@
...
+  class ClientService
+    require 'httparty'
+    require 'colorize'
+
+    USER_AGENT = 'Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) '\
+                'AppleWebKit/537.36 (KHTML, like Gecko) Chrome/91.0.4472.77 Safari/537.36'
+
+    def initialize(keyword, lang = 'en')
+      escaped_keyword = CGI.escape(keyword)
+      @uri = URI("https://google.com/search?q=#{escaped_keyword}&hl=#{lang}&gl=#{lang}")
```
junan: How about defining the base search url as constant like USER_AGENT?
```
BASE_SEARCH_URL = `https://google.com/search`
```
```suggestion
      @uri = URI("#{BASE_SEARCH_URL}?q=#{escaped_keyword}&hl=#{lang}&gl=#{lang}")
```

I think its a bit cleaner

spec/services/google_service/client_service_spec.rb
```diff
@@ -0,0 +1,19 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe GoogleService::ClientService, type: :service do
+  context 'when querying a simple keyword' do
+    it 'returns an HTTParty Response', vcr: 'google_search' do
```
junan:
```suggestion
    it 'returns a HTTParty Response', vcr: 'google_search' do
```

app/services/google_service/client_service.rb
malparty: Smart Rails!!

app/services/google_service/client_service.rb
malparty: Can do! ;-)

spec/services/google_service/client_service_spec.rb
malparty: @junan  `an` vs `a` is decided based on sounding (and ease of pronunciation), not based on vowel/consonant. So "an HTTP request" or "an hour" are correct ;-)

I had this issue last month ;-)

app/services/google_service/client_service.rb
malparty: Is call a real convention or just a part of the example? 
If it's a valid convention, I would submit a RFC in Compass so that we can add 1 or 2 lines to clarify what is convention from what is purely example :)

malparty: PS: Rename done, convention or not, it's much cleaner :D

config/routes.rb
```diff
@@ -1,11 +1,11 @@
...
   use_doorkeeper do
     skip_controllers :authorizations, :applications, :authorized_applications
   end
 
   devise_for :users
 
-  root to: 'keywords#index'
-
-  resources :keywords, only: :index
+  resources :keywords, only: [:index, :create]
```
ankitkalia1195: `create` action can be removed

app/services/google_service/client_service.rb
```diff
@@ -0,0 +1,40 @@
+# frozen_string_literal: true
+
+module GoogleService
+  class ClientService
+    USER_AGENT = 'Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) '\
+                'AppleWebKit/537.36 (KHTML, like Gecko) Chrome/91.0.4472.77 Safari/537.36'
+
+    BASE_SEARCH_URL2 = 'https://google.com/search'
+
+    def initialize(keyword, lang = 'en')
```
ankitkalia1195: https://nimblehq.co/compass/development/code-conventions/ruby/#methods

We dont always implement this this convention, but I guess here it could be applicable to use `keyword` arguments. IMO public methods which are called from outside the class and have multiple arguments passed are a good case for implementing keyword arguments.
```
  # Public method, use keyword argument.
  def initialize(keyword:, lang: 'en')
      escaped_keyword = CGI.escape(keyword)
      @uri = URI("#{BASE_SEARCH_URL2}?q=#{escaped_keyword}&hl=#{lang}&gl=#{lang}")
  end
```

app/services/google_service/client_service.rb
```diff
@@ -0,0 +1,40 @@
+# frozen_string_literal: true
+
+module GoogleService
```
ankitkalia1195: Can we remove `Service` from namespace, and only use `Google:ClientService`.

app/services/google_service/client_service.rb
```diff
@@ -0,0 +1,40 @@
...
+                'AppleWebKit/537.36 (KHTML, like Gecko) Chrome/91.0.4472.77 Safari/537.36'
+
+    BASE_SEARCH_URL2 = 'https://google.com/search'
+
+    def initialize(keyword, lang = 'en')
+      escaped_keyword = CGI.escape(keyword)
+      @uri = URI("#{BASE_SEARCH_URL2}?q=#{escaped_keyword}&hl=#{lang}&gl=#{lang}")
+    end
+
+    def call
```
ankitkalia1195: IMO with this implementation, what is becoming confusing is that value of instance variable `result` can be changed anywhere in any function, which can become difficult to track the value and reduces the readability of the code. Can you propose another implementation, maybe change the method name to `call!`, which can raise a error.

config/routes.rb
malparty: bien vu!!! (nice catch!)

app/services/google_service/client_service.rb
malparty: I'll use it as a rule of thumb :) 
I remember earlier in this project I faced this question ` when to use these or not ` so yep, I'll update that :)

app/services/google_service/client_service.rb
malparty: Yes we can :) 
That will solve my question "should I pluralize GoogleService[s] or not?" :)

app/services/google_service/client_service.rb
malparty: @ankitkalia1195 
I've try to set `result` as scoped variable, thus preventing it to be access somewhere else. 

For the `call!`, I'm not sure: The ClientService is built so that errors are managed. In case an issue occurs, the method returns nil. Maybe I can add some comments to document the service? Or is the bang (!) still needed? :) 

Thanks!

ankitkalia1195: @malparty The reason exception can be raised here is, because you would be calling this function somewhere inside a background worker(sidekiq). Exception would automatically cause the worker to fail, and job could be automatically retried. 

```
   class KeywordWorker
    
    def perform
      Google::ClientService.call
      # If above call raises exception, the job would fail and can be retried automatically
    end
```


With current approach, altough you are still logging the response, but the worker would not retry the job and job would be marked as success. Anyways you can revisit it when you implement that part. Other changes that could be done, if you decide that the function will not raise exception.

* Function returns false instead of `nil` when there is any error. (similary to active record save)
* validate_result should be before the exception handling

```
   result = HttpParty.get()
   return false unless valid_result?(result)
 
   result
   rescue =>e
     log
    false
```

malparty: @junan  / @ankitkalia1195 

At first I was afraid to put the `valid_result?` method inside the begin/rescue blocks: In the past I've been told we should not try/catch the whole code, only what is likely to throw errors (like the HTTP request -- as it depends on an external system). 

But if you tell me it's OK with Rails, then awesome! 

I love the eloquence of `return false unless valid_result?(result)` ! :) 

though I have a last question: What is the difference between returning `nil` and returning `false` ? 
As if I remember, doing `if nil` is the same as `if false`, no? :) 

Thanks for your info!

malparty: I keep this conversation opened: Code changed a lot, I've added more specs to test. You might have related comments to add in here :)

Gemfile
```diff
@@ -13,6 +13,7 @@ gem 'fabrication' # Fabrication generates objects in Ruby. Fabricators are schem
 gem 'sidekiq' # background processing for Ruby
 gem 'bootsnap', require: false # Reduces boot times through caching; required in config/boot.rb
 gem 'i18n-js', '3.5.1' # A library to provide the I18n translations on the Javascript
+gem 'httparty' #Makes http fun again! Ain't no party like a httparty, because a httparty don't stop.
```
junan: Can we make the comment a bit simple and more introductory?
```suggestion
gem 'httparty' # A HTTP client for Ruby.
```

app/services/google_service/client_service.rb
junan: > What is the difference between returns nil and returning false?


We return a boolean value(`true`, `false`) when the method ends with `?` like `valid?` which expects a boolean value. for other method (without `?` end) I personally return `true/false` to ensure the method's success or  failure.

> As if I remember, doing if nil is the same as if false, no? :)

In this case they are same though they are different data type. Like `nil == false` will return `false`

spec/services/google/client_service_spec.rb
```diff
@@ -0,0 +1,36 @@
...
+
+    it 'queries Google Search', vcr: 'google_search' do
+      path = described_class.new(keyword: FFaker::Lorem.word).call.request.path
+
+      expect(path.to_s).to start_with('https://www.google.com/search')
+    end
+  end
+
+  context 'when google returns an HTTP error' do
+    it 'returns a nil result', vcr: 'google_warn' do
```
junan:
```suggestion
    it 'returns false', vcr: 'google_warn' do
```

spec/services/google/client_service_spec.rb
```diff
@@ -0,0 +1,36 @@
...
+    it 'returns an HTTParty Response', vcr: 'google_search' do
+      result = described_class.new(keyword: FFaker::Lorem.word).call
+
+      expect(result).to be_an_instance_of(HTTParty::Response)
+    end
+
+    it 'queries Google Search', vcr: 'google_search' do
+      path = described_class.new(keyword: FFaker::Lorem.word).call.request.path
+
+      expect(path.to_s).to start_with('https://www.google.com/search')
```
junan: I think it can be used `Google::ClientService.BASE_SEARCH_URL` instead of checking hard coded value
```suggestion
      expect(path.to_s).to start_with(Google::ClientService.BASE_SEARCH_URL)
```

spec/services/google/client_service_spec.rb
malparty: Oh yes! Nice!!

app/services/google/client_service.rb
```diff
@@ -0,0 +1,41 @@
...
+      false
+    end
+
+    private
+
+    # Inspect Http response status code
+    # Any non 200 response code will be logged
+    # response is set to nil in order to notify the error
+    def valid_result?(result)
+      return result if result&.response&.code == '200'
```
ankitkalia1195: we can return `true` here is this method is predicate method

```
    return true if result&.response&.code == '200'
```

spec/services/google/client_service_spec.rb
```diff
@@ -0,0 +1,36 @@
...
+    end
+  end
+
+  context 'when google returns an HTTP error' do
+    it 'returns false', vcr: 'google_warn' do
+      result = described_class.new(keyword: FFaker::Lorem.word).call
+
+      expect(result).to be_falsey
```
ankitkalia1195: we can compare exact value, falsey means either of `nil or false`(which are falsy values in ruby)

```
   expect(result).to eq(false)
```

app/services/google/client_service.rb
```diff
@@ -0,0 +1,41 @@
...
+      Rails.logger.error "Error: Query Google with '#{@escaped_keyword}' thrown an error: #{e}".colorize(:red)
+
+      false
+    end
+
+    private
+
+    # Inspect Http response status code
+    # Any non 200 response code will be logged
+    # response is set to nil in order to notify the error
```
ankitkalia1195: This comment can be removed
```
  response is set to nil in order to notify the error
```

spec/services/google/client_service_spec.rb
malparty: Nice catch! @junan, it made me realized that google redirects to www.google.com and so I've updated the `BASE_SEARCH_URL` so that it includes the www.  ;-)

app/services/google/client_service.rb
malparty: True!

spec/services/google/client_service_spec.rb
malparty: Understood! Good to know! :)

app/services/google_service/client_service.rb
malparty: Thanks, very clear explanation!! :)

ankitkalia1195 (approved):

junan (approved):

PR #39, opened 2021-06-18
[#7] [Backend] As a User, I can query a single keyword and get its Google search results parsed

### LINKED ISSUE #7: [Backend] As a User, I can query a single keyword and get its Google search results parsed
## Why?
In order to see results directly and save time/effort.

### Additionnal notes:
Problems to solve:
- Save time for user by parsing Google search HTML content automatically 
- Enabling advanced results search for later features

No Gos:
- No data persistence [yet]
- No Csv File upload/parsing [yet]
- No fancy UI as such UI would not be sustained

## Who Benefits?
Logged users.

malparty:
- #7 

## What happened

Parse the HTML content from previously done search queries.

## Insight

- Open `rails console`
- Create a search (from previous PR feature): `result = GoogleService::ClientService.new('YOUR KEYWORD').query_result; nil` (note the final `; nil` is just here to avoid getting the html result in console output)
- Finally parse the search with `parsed = GoogleService::ParserService.new(result); nil`
- Now you can check all the parsed fields from `parsed` variable: ads_top_count, ads_page_count, ads_top_url, ads_page_url, non_ads_result_count, total_link_count

## Proof Of Work

Keyword used: `vpn`
image

Keyword used: `squarespace`

image


app/services/google/parser_service.rb
```diff
@@ -0,0 +1,66 @@
...
+      @document.css('#footcnt a').add_class('footer-links')
+    end
+
+    # Write parsed data directly in the keyword object argument
+    # and return back the keyword object
+    def parse_into!(keyword)
+      keyword.ads_top_count = ads_top_count
+      keyword.ads_page_count = ads_page_count
+      keyword.ads_top_url = ads_top_url
+      keyword.ads_page_url = ads_page_url
```
ankitkalia1195: @malparty Service classes usually have just a 2 public methods `call` and `initialize`. Also, IMO from this class it could be better to return the attributes in hash, instead of modify the keyword object.

spec/services/google/parser_service_spec.rb
```diff
@@ -0,0 +1,52 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe Google::ParserService, type: :service do
+  context 'when parsing a page having 1 top ad' do
+    it 'counts exactly 1 top ad', vcr: 'google_search_top_ads_1' do
```
ankitkalia1195: Can we simplify vcr casette names by using namespace

```
   google_search/top_ad_1
```

app/services/google/parser_service.rb
malparty: Nice, thanks to this feedback I discovered we can assign a hash values to ActiveRecord instance doing that:
```ruby
keyword.attributes = hash_of_attributes
``` 
I'll update the code accordingly ;-)

malparty: Attribute methods set as private, `parse_into!` renamed in `call`, specs updated accordingly :) 
It should be good now ;-)

ankitkalia1195 (approved):

app/services/google/parser_service.rb
```diff
@@ -0,0 +1,68 @@
...
+        ads_top_url: ads_top_url,
+        ads_page_url: ads_page_url,
+        non_ads_result_count: non_ads_result_count,
+        non_ads_url: non_ads_url,
+        total_link_count: total_link_count,
+        html: @html
+      }
+    end
+
+    private
```
junan, line 37: Related to https://github.com/malparty/google-search-ruby/pull/46#discussion_r660288299, I think we can do here also, defining `attr_reader` and use accessor while retrieving value
```suggestion
    private
    
    attr_reader :html, document
```

spec/services/google/client_service_spec.rb
```diff
@@ -4,27 +4,27 @@
 
 RSpec.describe Google::ClientService, type: :service do
```
junan, line 5: We are missing it, all the testing code should be inside this block
```ruby
describe '#call' do
...
...
end
```

spec/services/google/parser_service_spec.rb
```diff
@@ -0,0 +1,52 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe Google::ParserService, type: :service do
```
junan, line 5: Same as above comment

app/services/google/parser_service.rb
malparty: True, nice catch! 
Done ✅

spec/services/google/parser_service_spec.rb
malparty: Thanks!

spec/services/google/client_service_spec.rb
malparty: Yup! :)

junan (approved):

Issue #27, opened 2021-06-08
[Frontend] As a User, I can see a list of my previous keywords


malparty:
Build Front End for #26 

## Why?
In order for the user to navigate through his previously uploaded keywords. 
In order to provide an entry point for #8 & #12

### Additional notes:
Problems to solve:
- User can retrieve past data thanks to persistence
- User can see many keywords at once
- User can see results quickly even when many (10K +) keywords are already persisted (performance & UX to address)
- User can use his phone or desktop to navigate via UI

No Gos:
- No search feature [yet]
- No filter feature
- No multiple keywords upload/parsing [yet]
- No CSV File upload [yet]

## Who Benefits?
Logged Users.

PR #46, opened 2021-06-24
[#26] [BackEnd] As a User, I can see a list of my previous keywords

### LINKED ISSUE #26: [BackEnd] As a User, I can see a list of my previous keywords
## Why?

### Additional notes:
Problems to solve:

No Gos:
- No search feature [yet]
- No filter feature
- No CSV File upload [yet]

## Who Benefits?
Logged Users.

malparty:
- #26 

## What happened

- Create Keyword model
- Add seeds
- Add keyword#index Model query to list keywords (include pagination)

## Insight

- Create Keyword model
    - name has index
    - belongs to a user
- Add seeds
    - admin user (dev environment) in order to seed keywords to a user
    - 100x random keywords to this admin user
- Add keyword#index Model query to list keywords (include pagination)
    - using the Query Object pattern
        - only order by name asc (at the moment)
        - build to ease future extension 
    - use pagy gem for pagination

## Proof Of Work

Tests cover the `keywords_query` object:
image

UI tests will be added in the FrontEnd issue #27 and requests tests in the #25

### DIFF
--- a/app/controllers/keywords_controller.rb
+++ b/app/controllers/keywords_controller.rb
@@ -4,8 +4,10 @@ class KeywordsController < ApplicationController
   include Pagy::Backend
 
   def index
+    pagy, keywords = pagy(KeywordsQuery.new(current_user).call)
+
     render locals: {
-      keywords: pagy(KeywordsQuery.new(current_user).call)
+      pagy: pagy, keywords: keywords
     }
   end
 end
--- a/app/controllers/keywords_controller.rb
+++ b/app/controllers/keywords_controller.rb
@@ -4,10 +4,17 @@ class KeywordsController < ApplicationController
   include Pagy::Backend
 
   def index
-    pagy, keywords = pagy(KeywordsQuery.new(current_user).call)
+    pagy, keywords = pagy(keywords)
 
     render locals: {
       pagy: pagy, keywords: keywords
     }
   end
+
+  private
+
+  def keywords
+    KeywordsQuery.new(current_user).call
+  end
+
 end
--- a/app/controllers/keywords_controller.rb
+++ b/app/controllers/keywords_controller.rb
@@ -4,10 +4,10 @@ class KeywordsController < ApplicationController
   include Pagy::Backend
 
   def index
-    pagy, keywords = pagy(keywords)
+    pagy, keywords_list = pagy(keywords)
 
     render locals: {
-      pagy: pagy, keywords: keywords
+      pagy: pagy, keywords: keywords_list
     }
   end
 
--- a/app/controllers/keywords_controller.rb
+++ b/app/controllers/keywords_controller.rb
@@ -16,5 +16,4 @@ class KeywordsController < ApplicationController
   def keywords
     KeywordsQuery.new(current_user).call
   end
-
 end
--- a/app/controllers/keywords_controller.rb
+++ b/app/controllers/keywords_controller.rb
@@ -1,5 +1,11 @@
 # frozen_string_literal: true
 
 class KeywordsController < ApplicationController
-  def index; end
+  include Pagy::Backend
+
+  def index
+    render locals: {
+      keywords: pagy(KeywordsQuery.new(current_user).call)
+    }
+  end
 end
--- a/app/controllers/registrations_controller.rb
+++ b/app/controllers/registrations_controller.rb
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 
 class RegistrationsController < Devise::RegistrationsController
+  # Override user hard-delete (from Devise) with user soft-delete
   def destroy
     resource.discard
 
--- a/app/controllers/registrations_controller.rb
+++ b/app/controllers/registrations_controller.rb
@@ -1,0 +1,15 @@
+# frozen_string_literal: true
+
+class RegistrationsController < Devise::RegistrationsController
+  def destroy
+    resource.discard
+
+    Devise.sign_out_all_scopes ? sign_out : sign_out(resource_name)
+
+    set_flash_message :notice, :destroyed
+
+    yield resource if block_given?
+
+    respond_with_navigational(resource) { redirect_to after_sign_out_path_for(resource_name) }
+  end
+end
--- a/app/models/keyword.rb
+++ b/app/models/keyword.rb
@@ -1,7 +1,11 @@
 # frozen_string_literal: true
 
 class Keyword < ApplicationRecord
+  include Discard::Model
+
   validates :name, presence: true, length: { maximum: 255 }
 
   belongs_to :user, inverse_of: :keywords
+
+  default_scope -> { kept }
 end
--- a/app/models/keyword.rb
+++ b/app/models/keyword.rb
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
 class Keyword < ApplicationRecord
+  validates :name, presence: true, length: { maximum: 255 }
+
   belongs_to :user, inverse_of: :keywords
 end
--- a/app/models/keyword.rb
+++ b/app/models/keyword.rb
@@ -1,0 +1,5 @@
+# frozen_string_literal: true
+
+class Keyword < ApplicationRecord
+  belongs_to :user, inverse_of: :keywords
+end
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -1,7 +1,11 @@
 # frozen_string_literal: true
 
 class User < ApplicationRecord
-  has_many :keywords, inverse_of: :user, dependent: :delete_all
+  include Discard::Model
+
+  has_many :keywords, inverse_of: :user, dependent: :destroy
+
+  default_scope -> { kept }
 
   # Include default devise modules. Others available are:
   # :confirmable, :lockable, :timeoutable, :trackable and :omniauthable
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -1,6 +1,8 @@
 # frozen_string_literal: true
 
 class User < ApplicationRecord
+  has_many :keywords, inverse_of: :user, dependent: :delete_all
+
   # Include default devise modules. Others available are:
   # :confirmable, :lockable, :timeoutable, :trackable and :omniauthable
   devise :database_authenticatable, :registerable,
--- a/app/queries/keywords_query.rb
+++ b/app/queries/keywords_query.rb
@@ -8,12 +8,12 @@ class KeywordsQuery
   end
 
   def call
-    @keywords = order
+    @keywords = order_by_name
   end
 
   private
 
-  def order
+  def order_by_name
     keywords.order(:name)
   end
 end
--- a/app/queries/keywords_query.rb
+++ b/app/queries/keywords_query.rb
@@ -11,7 +11,9 @@ class KeywordsQuery
 
   private
 
+  attr_reader :keywords
+
   def order_by_name
-    @keywords.order(:name)
+    keywords.order(:name)
   end
 end
--- a/app/queries/keywords_query.rb
+++ b/app/queries/keywords_query.rb
@@ -1,8 +1,6 @@
 # frozen_string_literal: true
 
 class KeywordsQuery
-  attr_reader :keywords
-
   def initialize(user)
     @keywords = user.keywords
   end
@@ -14,6 +12,6 @@ class KeywordsQuery
   private
 
   def order_by_name
-    keywords.order(:name)
+    @keywords.order(:name)
   end
 end
--- a/app/queries/keywords_query.rb
+++ b/app/queries/keywords_query.rb
@@ -8,7 +8,7 @@ class KeywordsQuery
   end
 
   def call
-    @keywords = order_by_name
+    order_by_name
   end
 
   private
--- a/app/queries/keywords_query.rb
+++ b/app/queries/keywords_query.rb
@@ -1,0 +1,19 @@
+# frozen_string_literal: true
+
+class KeywordsQuery
+  attr_reader :keywords
+
+  def initialize(user)
+    @keywords = user.keywords
+  end
+
+  def call
+    @keywords = order
+  end
+
+  private
+
+  def order
+    keywords.order(:name)
+  end
+end
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -5,7 +5,7 @@ Rails.application.routes.draw do
     skip_controllers :authorizations, :authorized_applications, :tokens, :token_info
   end
 
-  devise_for :users
+  devise_for :users, :controllers => { :registrations => 'registrations' }
 
   resources :keywords, only: :index
 
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -5,7 +5,7 @@ Rails.application.routes.draw do
     skip_controllers :authorizations, :authorized_applications, :tokens, :token_info
   end
 
-  devise_for :users, :controllers => { :registrations => 'registrations' }
+  devise_for :users, controllers: { registrations: :registrations }
 
   resources :keywords, only: :index
 
--- a/db/migrate/20210624033123_create_keywords.rb
+++ b/db/migrate/20210624033123_create_keywords.rb
@@ -1,0 +1,19 @@
+class CreateKeywords < ActiveRecord::Migration[6.1]
+  def change
+    create_table :keywords do |t|
+      t.string :name, null: false
+
+      t.references :user, index: true, null: false
+
+      t.timestamps
+    end
+
+    add_index :keywords, :name
+
+    add_foreign_key(
+      :keywords,
+      :users,
+      column: :user_id
+    )
+  end
+end
--- a/db/migrate/20210624033123_create_keywords.rb
+++ b/db/migrate/20210624033123_create_keywords.rb
@@ -1,19 +1,15 @@
+# frozen_string_literal: true
+
 class CreateKeywords < ActiveRecord::Migration[6.1]
   def change
     create_table :keywords do |t|
       t.string :name, null: false
 
-      t.references :user, index: true, null: false
+      t.references :user, index: true, null: false, foreign_key: true
 
       t.timestamps
     end
 
     add_index :keywords, :name
-
-    add_foreign_key(
-      :keywords,
-      :users,
-      column: :user_id
-    )
   end
 end
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -16,5 +16,5 @@ end
 if Rails.env.development?
   user = User.where(email: 'admin@nimblehq.co').first || Fabricate(:admin, email: 'admin@nimblehq.co')
 
-  100.times { Fabricate(:keyword, user: user) }
+  Fabricate.times(100, :keyword, user: user)
 end
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -14,7 +14,7 @@ if Doorkeeper::Application.count.zero?
 end
 
 if Rails.env.development?
-  user = User.where(email: 'admin@nimblehq.co').first || Fabricate(:admin, email: 'admin@nimblehq.co')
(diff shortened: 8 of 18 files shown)


app/controllers/keywords_controller.rb
```diff
@@ -1,5 +1,13 @@
 # frozen_string_literal: true
 
 class KeywordsController < ApplicationController
-  def index; end
+  include Pagy::Backend
+
+  def index
+    pagy, keywords = pagy(KeywordsQuery.new(current_user).call)
```
junan: How about separate the KeywordQuery stuff into a separate method like we did in our [previous project](https://github.com/nimblehq/the1-web/blob/f97fcb48e50e605a92b35398dbd8e3ec6324a990/engines/the1_article/app/controllers/the1_article/articles_controller.rb#L15). I will be clean

```suggestion
    pagy, keywords = pagy(keywords).call)
```


```Ruby
private

def keywords
   KeywordsQuery.new(current_user).call
end
```

app/models/keyword.rb
```diff
@@ -0,0 +1,5 @@
+# frozen_string_literal: true
+
+class Keyword < ApplicationRecord
```
junan, line 3: We can add model level validation
```suggestion
class Keyword < ApplicationRecord

validates: name, presence: true
```

app/models/user.rb
```diff
@@ -1,6 +1,8 @@
 # frozen_string_literal: true
 
 class User < ApplicationRecord
+  has_many :keywords, inverse_of: :user, dependent: :delete_all
```
ankitkalia1195: Usually `dependent: :destroy` is more commonly used

app/models/keyword.rb
```diff
@@ -0,0 +1,5 @@
...
+  belongs_to :user, inverse_of: :keywords
```
ankitkalia1195, line 8: According to our compass conventions, should we use soft delete here. 
Screen Shot 2564-06-25 at 15 45 16

```
   Always soft-delete relational data for integrity and audit purposes. Gems such as paranoia and discard provide ready-to-use solutions.
```

app/queries/keywords_query.rb
```diff
@@ -0,0 +1,19 @@
...
+    @keywords = user.keywords
+  end
+
+  def call
+    @keywords = order
+  end
+
+  private
+
+  def order
```
ankitkalia1195: can be rename it to order_by_name. or even keeping it in same call function can also be ok for now.

db/migrate/20210624033123_create_keywords.rb
```diff
@@ -0,0 +1,19 @@
+class CreateKeywords < ActiveRecord::Migration[6.1]
+  def change
+    create_table :keywords do |t|
+      t.string :name, null: false
+
+      t.references :user, index: true, null: false
```
ankitkalia1195: does adding `foreign_key: true` here work?

db/seeds.rb
```diff
@@ -12,3 +12,9 @@
   Doorkeeper::Application.create(name: "iOS client", redirect_uri: "", scopes: "")
   Doorkeeper::Application.create(name: "Android client", redirect_uri: "", scopes: "")
 end
+
+if Rails.env.development?
+  user = User.where(email: 'admin@nimblehq.co').first || Fabricate(:admin, email: 'admin@nimblehq.co')
```
junan: Using `exists?` would be more simple and performant 
```suggestion
  user = User.exists?(email: "admin@nimblehq.co") || Fabricate(:admin, email: 'admin@nimblehq.co')
```

spec/models/keyword_spec.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe Keyword, type: :model do
+  pending "add some examples to (or delete) #{__FILE__}"
```
junan: We can add `name` attribute validation test like https://github.com/nimblehq/the1-web/blob/f97fcb48e50e605a92b35398dbd8e3ec6324a990/spec/models/the1_contact/partner_spec.rb#L7

app/controllers/keywords_controller.rb
malparty: Nice!! I'll do that :)

malparty: Fixed ;-)

app/models/keyword.rb
malparty: Thanks @junan :)
I've read that Rails won't raise an error if a keyword is bigger than 255 chars (and the string would simply be truncated). Is it a common practice to add also a max length validate? :)

malparty: Okay, I've added:
```ruby
validates :name, presence: true, length: { maximum: 255 }
```

as well as added model tests to secure these 2x validation rules :) 
@junan  I keep the conversation open to hear from the previous comment :) Thanks!

app/models/keyword.rb
malparty: Yes, makes perfect sense. 

As there will be no delete/destroy action exposed for keywords, I will try to handle it on the only scenario when a user "cancel his account" (so that it soft-delete the user, but it can keep keywords as is).

Would that be ok @ankitkalia1195 ? :) Thanks!

spec/models/keyword_spec.rb
malparty: Nice example :) I've added the expect presence of name for keyword too :) 
Thanks!

db/seeds.rb
malparty: Good feedback as it reminded me about the `exist?` method :) :) 

In this scenario I expected to assign 100 keywords on the admin user, **and** create this user if not done yet.

If I use exist? it will return `true` and assign `true` to user, isn't it? 
What do you think @junan ? :)

db/migrate/20210624033123_create_keywords.rb
malparty: Tested: Yes, it works well! :) 
Code updated ;-)

spec/models/keyword_spec.rb
```diff
@@ -0,0 +1,31 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe Keyword, type: :model do
+  describe 'Validations' do
+    it { is_expected.to validate_presence_of(:name) }
```
ankitkalia1195, line 7: Usually if a validation can be covered by shoulda matchers, we dont need to test it manually.

app/models/keyword.rb
junan: > I've read that Rails won't raise an error if a keyword is bigger than 255 chars (and the string would simply be truncated). Is it a common practice to add also a max length validate? :)

It is up to you I think, ` length: { maximum: 255 }` one better addition I  also think. I usually use `string` column for shorter  text and goes for `text`(it can store long text) column for longer text

spec/models/keyword_spec.rb
```diff
@@ -0,0 +1,31 @@
...
+
+    context 'given a blank name' do
+      it 'raises a RecordInvalid error' do
+        expect { Fabricate(:keyword, name: " \n") }.to raise_error(ActiveRecord::RecordInvalid)
+      end
+    end
+
+    context 'given a too long name' do
```
ankitkalia1195: Can we do this via 
```
   it { is_expected.to validate_length_of(:name).is_greater_than(5).is_at_most(19)
```

https://github.com/thoughtbot/shoulda-matchers/blob/master/lib/shoulda/matchers/active_model/validate_length_of_matcher.rb

app/models/keyword.rb
malparty: ok, soft delete for Users implemented and tested. 
Do you think we need to add tests @ankitkalia1195 ? 
Candidates for tests:
- Once deleted, a user get logged out
- Once deleted, a user cannot log in anymore
- Once deleted, a user cannot reset his password

it's build-in with devise and discard, so I'm not sure if tests are needed in here~
Thanks for your advises :)

spec/models/keyword_spec.rb
malparty: YAD! (Yet Another Discovery)
Thanks! :) 

Fixed~

db/seeds.rb
junan: Ahh, ya. Its return `true` but you need to add a user.

Alternatively, you can use `first_or_create` method also like bellow like bellow. This is optional.
```ruby
 user = User.where(email: 'admin@nimblehq.co').first_or_create(Fabricate.attributes_for(:admin).except(:email))
 ```

File: app/controllers/registrations_controller.rb [hunk 11]
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+
+class RegistrationsController < Devise::RegistrationsController
+  def destroy
```
junan, line 5: This PR is only for listing keyword, not sure why we need it 🤔

app/queries/keywords_query.rb
```diff
@@ -0,0 +1,19 @@
...
+
+class KeywordsQuery
+  attr_reader :keywords
+
+  def initialize(user)
+    @keywords = user.keywords
+  end
+
+  def call
+    @keywords = order_by_name
```
junan: I guess, we dont need to assign here, because we will not later access `keywords` from `KeywordsQuery.new(current_user).keywords`. What do you think?
```suggestion
    order_by_name
```

config/routes.rb
```diff
@@ -5,7 +5,7 @@
     skip_controllers :authorizations, :authorized_applications, :tokens, :token_info
   end
 
-  devise_for :users
+  devise_for :users, :controllers => { :registrations => 'registrations' }
```
junan:
```suggestion
  devise_for :users, controllers: { registrations: :registrations }
```

db/seeds.rb
malparty: Interesting, it prevent to think about the logic behind `||` ;-) I'll take that then :)  
Done~

app/controllers/registrations_controller.rb
malparty: Adding the relation `has_many` brought the lights on "how do we delete keywords". The solution making more sense according to me was not to delete any keywords, but to delete users if needed. It was also the occasion to practice with soft-delete (without implementing a non-asked feature to delete keywords) :) 
But yeah, it's arguable that it should have been done earlier in the authentication PRs.

config/routes.rb
malparty: I got trapped by devise tutorial!!! I should not trust in copy/paste anymore with Ruby :P

malparty: Fixed.

app/queries/keywords_query.rb
malparty: true :) ✅
Fixed!

app/controllers/registrations_controller.rb
malparty: let me know if you think I should extract it into a chore branch~ @junan

junan: Ok, its fine to me as it is related to current soft delete changes and changes is not big

app/queries/keywords_query.rb
```diff
@@ -0,0 +1,19 @@
+# frozen_string_literal: true
+
+class KeywordsQuery
+  attr_reader :keywords
```
junan: I think it should be private, as we are not accessing the keyword like  `KeywordsQuery.new(User.new).keywords`

```ruby
...
private

attr_reader :keywords
...
```

app/models/keyword.rb
ankitkalia1195: @malparty It seems you have again overcomplicated things, but its okay. For testing soft delete in models, you could have just tested them via rails console. We add soft delete while designing the model class, does not necessary mean that we should have some UI to test it.

ankitkalia1195: The main motive is to just ensure that data does not get deleted from database, whether is user or keyword. So usually when working on projects, we add soft delete to models straight away. Even though there might not be a way for users to delete it directly

app/queries/keywords_query.rb
malparty: But then we do not need the attribute at all, no? :) 
(I've added the attribute as it was part of an example of a QueryObject pattern -- but indeed, if we do not need it, there is no point to have it).
I've updated the code to remove the attr, let me know if it's ok @junan  :)

app/models/keyword.rb
ankitkalia1195: IMO we can keep using paranoia instead of discard, as it is in the rails template. With paranoia, we dont need to override anything like `devise` as well. Just calling `user.delete` will trigger soft delete

ankitkalia1195: For reference, you can check project mia. https://github.com/nimblehq/mia-web/tree/development/app/models. Here we have `acts_as_paranoid` in all models

malparty: yes, the idea was to practice, there is a high chance that in a future project I get the need for a user soft-delete feature  ^^, 
For the choice of `discard` over `paranoia`, it's due to the notice heading `paranoia` readme: 
> paranoia has some surprising behaviour (like overriding ActiveRecord's delete and destroy) and is not recommended for new projects. See discard's README for more details.

I'll remember to add a soft-delete even when not needed -- with the intend to prevent unexpected hard-delete into future features.

app/queries/keywords_query.rb
junan: @malparty `attr_reader` actually  generates  a `getter`  method behind the scene. When we read data, we dont directly call the instance variable(ex: @keywords), rather we retrieve the data with a getter method (ex: `keywords` method which generated by `attr_reader :keywords`).

For example this class
```Ruby
class SomeClass
  # Define getter method in public if we want to access a, b, c outside of the class
  attr_reader :a, :b, :c
  
  def initialize(a, b, c)
    @a = a
    @b = b
    @c = c
  end

  def call
    # access with a/b/c using getter method instead of accessing @a/@b/@c directly
  end
  
  private
  
  # Define getter method in private if we dont want to access a, b, c outside of the class
  attr_reader :a, :b, :c

  def some_private_method
    # access with a/b/c using getter method instead of accessing @a/@b/@c directly
  end
end
```

So your class will be like this

```Ruby
class KeywordsQuery
  def initialize(user)
    @keywords = user.keywords
  end

  def call
    order_by_name
  end

  private

  attr_reader: keywords
  
  def order_by_name
    keywords.order(:name)
  end
end
```

app/models/keyword.rb
ankitkalia1195: @malparty Sure, we can use `discard` as well, altough in most of our projects we still use `paranoia`.

app/models/user.rb
```diff
@@ -1,6 +1,12 @@
 # frozen_string_literal: true
 
 class User < ApplicationRecord
+  include Discard::Model
+
+  has_many :keywords, inverse_of: :user, dependent: :destroy
```
ankitkalia1195, line 6: If we are using discard, then its better not to have `dependent: :destroy`, as it would hard delete the keywords. A better approach would be 

```
   class User
     includes Discard:Model

     after_discard do
       keywords.discard_all
    end
end
```

app/models/keyword.rb
ankitkalia1195: Basically `paranoia` is easier to use and does everything `magically`, `discard` is more explicit, and you have to write more code in order to achieve the result. In our rails template we still prefer `paranoia`, but you can bring it to a vote if it should be `discard`

malparty: Ok, yes I saw `paranoia` replace the destroy/delete methods so it has this "magic rails" feel :) 
As I started with `discard` and as you said, I already complicated things, then I've just added the soft delete to keywords -- as per convention.

| Some little rails console test: | It fills the discarded_at date: | 
|---|---|
| image | image |

And as the scope of keywords is `kept`, it does not display the keywords anymore:
image

app/queries/keywords_query.rb
malparty: Thanks for the example, though I'm not sure to understand the 'why'. 

So you mean any single private instance variable should never be accessed directly, but always via accessors? 
Is it to keep the flexibility of overriding the `getter` method later? Or any other reason? 

I'm asking to be sure I understand the why -- back in C# I used to utilize getter/setters solely for public instance variables (called "Properties"). Here the why was specific to maintain a stable contract (API) so that consumers of the class won't have to change when the variable does (as getter will maintain a bridge). But for private variables, I'm not sure to see exactly what we want to tackle~ 

Though I have already refactored the code to use the accessor ;-) 

Thanks @junan :)

app/models/user.rb
malparty: I'm not sure to agree 100% ^^
--> As you said, I've implemented the soft-delete on keywords in "anticipation for later need". 
Currently, there is absolutely 0 use case where we want to discard a keyword. 

Assuming we were in the future and had a UI to discard keywords: discarding all keywords when discarding a user would be wrong. If the user undiscard his account later, he would not be able to restore the state of discarded keywords (as all would be discarded). Plus the state of the keyword should not depend on the state of the user IMO.

The `dependent` is more here as a convention/best-practice, but because our code does not delete (nor destroy) any user, so it should not be triggered. 

What do you think? :) 

-- 
After a second reading, I think I understood the idea: You were more using `paranoia` as a "delete safety" (because it override delete/destroy), so that no developer would risk to accidentally delete a DB record, right? (in which case `discard` does not make the job as it is adding a method and still provide direct access to delete/destroy). And so... makes sense to setup paranoia by default even without real need for soft-delete ; but it maybe makes less sense to setup discard by default... Not sure if that's the point?

ankitkalia1195: @malparty Yes, thats why I recommend using paranoia as we are using it in most of our projects and it is present in our rails template. If you want to change the convention, I guess you can bring it to vote in web chapter. If you use paranoia, we can just add it to all models as safety with just one line.

ankitkalia1195: And yea to summarize, it makes better sense to setup paranoia by default on all models, than to setup discard on all models

malparty: Ok clear! :) Thanks for all the explanations~ 
Well, I do not personally want to change our standards if they work well ;-) 
But it can be interesting indeed to try to understand why paranoia does not recommend itself for new projects. I'll call for a vote, but I'm perfectly fine with paranoia if you and others had no issue with it ;-)

For this IC, if you don't mind I'd be very happy to keep it as-is and to move forward :D

app/controllers/registrations_controller.rb
```diff
@@ -0,0 +1,15 @@
...
+    resource.discard
+
+    Devise.sign_out_all_scopes ? sign_out : sign_out(resource_name)
+
```
ankitkalia1195, line 9: @malparty Lets add a comment as to why we are overriding this method.

ankitkalia1195 (approved):

app/queries/keywords_query.rb
junan: @malparty Theoretically we can access the instance variable directly inside the class, but we generally [access](https://github.com/nimblehq/the1-web/blob/development/engines/the1_auth/app/forms/the1_auth/token_form.rb) them with accessors. I see the following benefit accessing value with accessor methods rather than directly accessing them with instance variable  
- Defining `attr_reader` clearly tells us that those attributes are only for reading(not writing), so more clarity
- If we need to access those variables outside of class, we cant access them directly with the instance variable. We need to define the accessor. So It is better to use the accessor method always to access data, so more consistency
- Every time we access the variable value inside the class, we type one less character(`@`), so less typing

junan (approved):

app/queries/keywords_query.rb
malparty: Thanks, very clear and I did not think of the attr-reader only situation!!! Cool! :)

PR #47, opened 2021-06-24
[#27] [Frontend] As a User, I can see a list of my previous keywords

### LINKED ISSUE #27: [Frontend] As a User, I can see a list of my previous keywords

malparty:
- #27 

## What happened

Create views to list keywords (from database data provided by #26)

## Insight

- Using decorator to address keyword grouping logic
- Using `pagy` gem for pagination (with included bootstrap helpers)
- 1 scss component create `list-keywords-group`

## Proof Of Work

| Desktop | Mobile |
|---|---|
| image | image |

### DIFF
--- a/app/controllers/keywords_controller.rb
+++ b/app/controllers/keywords_controller.rb
@@ -7,7 +7,7 @@ class KeywordsController < ApplicationController
     pagy, keywords_list = pagy(keywords)
 
     render locals: {
-      pagy: pagy, keywords: keywords_list
+      pagy: pagy, keywords: KeywordsDecorator.new(keywords_list)
     }
   end
 
--- a/app/controllers/keywords_controller.rb
+++ b/app/controllers/keywords_controller.rb
@@ -7,7 +7,7 @@ class KeywordsController < ApplicationController
     pagy, keywords_list = pagy(keywords)
 
     render locals: {
-      pagy: pagy, keywords: KeywordsDecorator.new(keywords_list)
+      pagy: pagy, keywords: KeywordsPresenter.new(keywords_list)
     }
   end
 
--- a/app/controllers/keywords_controller.rb
+++ b/app/controllers/keywords_controller.rb
@@ -7,7 +7,7 @@ class KeywordsController < ApplicationController
     pagy, keywords_list = pagy(keywords)
 
     render locals: {
-      pagy: pagy, keywords: KeywordsPresenter.new(keywords_list)
+      pagy: pagy, keywords: KeywordsCollectionPresenter.new(keywords_list)
     }
   end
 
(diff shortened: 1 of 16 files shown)


app/controllers/keywords_controller.rb
```diff
@@ -7,7 +7,7 @@ def index
     pagy, keywords_list = pagy(keywords)
 
     render locals: {
-      pagy: pagy, keywords: keywords_list
+      pagy: pagy, keywords: KeywordsDecorator.new(keywords_list)
```
junan: Based on our conventions, decorator pattern should be only used for the backend and presenter pattern should be used in view. So I think it will be presenter instead of decorator as we are using it in the view

```Ruby
pagy: pagy, keywords_collection_presenter: KeywordsCollectionPresenter.new(keywords)
```

For reference, you can check this [controller](https://github.com/nimblehq/mia-web/blob/development/engines/mia_campaign/app/controllers/mia_campaign/campaign_influencers_controller.rb#L13) and [presenter](https://github.com/nimblehq/mia-web/blob/development/engines/mia_admin/app/presenters/mia_admin/withdrawals/campaign_influencer_collection_presenter.rb) in MIA project


Screen Shot 2021-06-29 at 10 28 54 AM

app/views/keywords/_list_keywords.html.erb
```diff
@@ -0,0 +1,12 @@
+<section class="list-keywords">
```
junan:
```suggestion
<section class="list-keyword">
```

app/controllers/keywords_controller.rb
malparty: Nice catch! I did not pay attention at the small warning in decorator as I thought `SimpleDelegator` was something totally separated from the Decorators part. 
Code updated & pushed @junan  :)

ankitkalia1195 (approved):

junan:
@malparty Many runrelated changes, need to resolve conflict 👍

app/presenters/keywords_collection_presenter.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+class KeywordsCollectionPresenter
```
junan, line 3: Related to this comment https://github.com/malparty/google-search-ruby/pull/46#discussion_r660288299, how about do the same also for this class?

```ruby
class KeywordsCollectionPresenter
  def initialize(keywords)
    @keywords = keywords
  end

  def groups
    keywords.group_by { |keyword| keyword.name[0].upcase.to_sym }
  end
  
  private
  
  attr_reader :keywords
end
```

malparty: Makes perfect sense! :)

junan (approved):

PR #49, opened 2021-06-25
[#25] [API] As a User, I can see a list of my previous keywords

### LINKED ISSUE #25: [API] As a User, I can see a list of my previous keywords
Build API for #26 

## Why?

### Additional notes:
Problems to solve:

No Gos:
- No search feature [yet]
- No filter feature
- No CSV File upload [yet]

## Who Benefits?
Logged Users.

malparty:
- #25

## What happened

Expose keywords list to API

## Insight

Basic keyword list (name/id only).
- Pagination implemented with Pagy
- Pagination made JSON::API compliant with this concern: `Pagy::JSONAPIConcern`
- API Endpoints [documented here](https://documenter.getpostman.com/view/16135891/TzeTKqCb)

## Proof Of Work

| Returns data as an array. Include pagination. | Error page overflow / Error invalid page |
|---|---|
| image | image | 
| image | image |

### DIFF
--- a/app/controllers/api/v1/keywords_controller.rb
+++ b/app/controllers/api/v1/keywords_controller.rb
@@ -9,11 +9,7 @@ module API
       def index
         pagy, keywords_list = pagy(keywords)
 
-        if keywords.any?
-          render json: KeywordSerializer.new(keywords_list, pagy_options(pagy))
-        else
-          render_empty I18n.t('keywords.empty_list')
-        end
+        render json: KeywordSerializer.new(keywords_list, pagy_options(pagy))
       end
 
       private
--- a/app/controllers/api/v1/keywords_controller.rb
+++ b/app/controllers/api/v1/keywords_controller.rb
@@ -10,7 +10,7 @@ module API
         pagy, keywords_list = pagy(keywords)
 
         if keywords.any?
-          render json: KeywordSerializer.new(keywords_list, pagy_options(pagy)).serializable_hash, status: :ok
+          render json: KeywordSerializer.new(keywords_list, pagy_options(pagy)).serializable_hash
         else
           render_empty
         end
@@ -19,7 +19,7 @@ module API
       private
 
       def render_empty
-        render json: { meta: I18n.t('keywords.empty_list'), data: [] }, status: :ok
+        render json: { meta: I18n.t('keywords.empty_list'), data: [] }
       end
 
       def keywords
--- a/app/controllers/api/v1/keywords_controller.rb
+++ b/app/controllers/api/v1/keywords_controller.rb
@@ -1,0 +1,35 @@
+# frozen_string_literal: true
+
+module API
+  module V1
+    class KeywordsController < ApplicationController
+      include ::Pagy::Backend
+      include API::V1::Pagy::JSONAPIConcern
+      include ErrorHandlerConcern
+
+      def index
+        pagy, keywords = pagy(KeywordsQuery.new(current_user).call)
+
+        if keywords.any?
+          render json: KeywordSerializer.new(keywords, pagy_options(pagy)).serializable_hash, status: :ok
+        else
+          render_empty
+        end
+      rescue ::Pagy::OverflowError => e
+        rescue_page_error("#{I18n.t('pagy.errors.overflow')} #{e.pagy.last}", e)
+      rescue ::Pagy::VariableError => e
+        rescue_page_error(I18n.t('pagy.errors.variable'), e)
+      end
+
+      private
+
+      def rescue_page_error(message, error)
+        render_error(details: message, source: error.variable, status: :unprocessable_entity)
+      end
+
+      def render_empty
+        render json: { meta: I18n.t('keywords.empty_list'), data: [] }, status: :ok
+      end
+    end
+  end
+end
--- a/app/controllers/api/v1/keywords_controller.rb
+++ b/app/controllers/api/v1/keywords_controller.rb
@@ -12,15 +12,11 @@ module API
         if keywords.any?
           render json: KeywordSerializer.new(keywords_list, pagy_options(pagy)).serializable_hash
         else
-          render_empty
+          render_empty I18n.t('keywords.empty_list')
         end
       end
 
       private
-
-      def render_empty
-        render json: { meta: I18n.t('keywords.empty_list'), data: [] }
-      end
 
       def keywords
         KeywordsQuery.new(current_user).call
--- a/app/controllers/api/v1/keywords_controller.rb
+++ b/app/controllers/api/v1/keywords_controller.rb
@@ -5,7 +5,6 @@ module API
     class KeywordsController < ApplicationController
       include ::Pagy::Backend
       include API::V1::Pagy::JSONAPIConcern
-      include ErrorHandlerConcern
 
       def index
         pagy, keywords_list = pagy(keywords)
@@ -15,17 +14,9 @@ module API
         else
           render_empty
         end
-      rescue ::Pagy::OverflowError => e
-        rescue_page_error("#{I18n.t('pagy.errors.overflow')} #{e.pagy.last}", e)
-      rescue ::Pagy::VariableError => e
-        rescue_page_error(I18n.t('pagy.errors.variable'), e)
       end
 
       private
-
-      def rescue_page_error(message, error)
-        render_error(details: message, source: error.variable, status: :unprocessable_entity)
-      end
 
       def render_empty
         render json: { meta: I18n.t('keywords.empty_list'), data: [] }, status: :ok
--- a/app/controllers/api/v1/keywords_controller.rb
+++ b/app/controllers/api/v1/keywords_controller.rb
@@ -10,7 +10,7 @@ module API
         pagy, keywords_list = pagy(keywords)
 
         if keywords.any?
-          render json: KeywordSerializer.new(keywords_list, pagy_options(pagy)).serializable_hash
+          render json: KeywordSerializer.new(keywords_list, pagy_options(pagy))
         else
           render_empty I18n.t('keywords.empty_list')
         end
--- a/app/controllers/api/v1/keywords_controller.rb
+++ b/app/controllers/api/v1/keywords_controller.rb
@@ -8,10 +8,10 @@ module API
       include ErrorHandlerConcern
 
       def index
-        pagy, keywords = pagy(KeywordsQuery.new(current_user).call)
+        pagy, keywords_list = pagy(keywords)
 
         if keywords.any?
-          render json: KeywordSerializer.new(keywords, pagy_options(pagy)).serializable_hash, status: :ok
+          render json: KeywordSerializer.new(keywords_list, pagy_options(pagy)).serializable_hash, status: :ok
         else
           render_empty
         end
@@ -29,6 +29,10 @@ module API
 
       def render_empty
         render json: { meta: I18n.t('keywords.empty_list'), data: [] }, status: :ok
+      end
+
+      def keywords
+        KeywordsQuery.new(current_user).call
       end
     end
   end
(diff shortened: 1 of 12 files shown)


app/controllers/api/v1/keywords_controller.rb
```diff
@@ -0,0 +1,39 @@
...
+
+      def index
+        pagy, keywords_list = pagy(keywords)
+
+        if keywords.any?
+          render json: KeywordSerializer.new(keywords_list, pagy_options(pagy)).serializable_hash, status: :ok
+        else
+          render_empty
+        end
+      rescue ::Pagy::OverflowError => e
```
ankitkalia1195: since this could be common error among several resources, it could good to add it into application_controller

```
    rescue_from Pagy::OverflowError, ::Pagy::VariableError => e
        render_error(details: message, source: error.variable, status: :unprocessable_entity)
    end
```

app/controllers/api/v1/keywords_controller.rb
```diff
@@ -0,0 +1,39 @@
...
+  module V1
+    class KeywordsController < ApplicationController
+      include ::Pagy::Backend
+      include API::V1::Pagy::JSONAPIConcern
+      include ErrorHandlerConcern
+
+      def index
+        pagy, keywords_list = pagy(keywords)
+
+        if keywords.any?
```
ankitkalia1195: what happens when you pass empty keywords_list, is the empty case really needed?

spec/requests/api/v1/keywords_controller_spec.rb
```diff
@@ -0,0 +1,94 @@
...
+        create_token_header(user)
+
+        get :index
+
+        expect(JSON.parse(response.body)['links'].keys).to contain_exactly('self', 'first', 'prev', 'next', 'last')
+      end
+    end
+
+    context 'given a valid page param' do
+      it 'returns the keywords to that page' do
```
ankitkalia1195: `returns the keywords for that page`

spec/requests/api/v1/keywords_controller_spec.rb
```diff
@@ -0,0 +1,94 @@
...
+        create_token_header(user)
+
+        get :index, params: { page: 2 }
+
+        expect(JSON.parse(response.body)['data'].count).to eq(1)
+      end
+    end
+
+    context 'given an invalid page param' do
+      it 'returns a 422 Unprocessable Entity' do
```
ankitkalia1195: `returns a 422 Unprocessable Entity`

app/controllers/api/v1/keywords_controller.rb
malparty: Nice! :) 
Done ;-)

spec/requests/api/v1/keywords_controller_spec.rb
malparty: I guess you mean
`returns a 422 Unprocessable Entity status code`
right? :) 
If yes, fixed!

app/controllers/api/v1/keywords_controller.rb
malparty: It's not changing much, the idea was to have a custom message.

Both are perfectly well working. 
I'm ok to remove it as the pros/cons are very small.

With this custom message:
- Lighter result
- Custom message can be reused by API consumer
```json
{
    "meta": "You do not have any keyword yet.",
    "data": []
}
```

Without this message:
- Format is more consistent between empty/not_empty queries
- No custom message
```json
{
    "data": [],
    "meta": {
        "total_pages": 1
    },
    "links": {
        "first": "http://localhost:3000/api/v1/keywords?page=1",
        "self": "http://localhost:3000/api/v1/keywords?page=1",
        "next": null,
        "prev": null,
        "last": "http://localhost:3000/api/v1/keywords?page=1"
    }
}
```

What do you think, @ankitkalia1195 ? :)

spec/requests/api/v1/keywords_controller_spec.rb
```diff
@@ -0,0 +1,94 @@
...
+  describe '#index' do
+    context 'given an unauthenticated user' do
+      it 'returns a 401 status code' do
+        get :index
+
+        expect(response.status).to eq(401)
+      end
+    end
+
+    context 'when no keywords' do
```
ankitkalia1195: can we change it to 'when there are no keywords', or 'when keyword_list is empty'?

malparty: Sure! :) Done ;-)

app/controllers/api/v1/keywords_controller.rb
```diff
@@ -0,0 +1,30 @@
...
+  module V1
+    class KeywordsController < ApplicationController
+      include ::Pagy::Backend
+      include API::V1::Pagy::JSONAPIConcern
+
+      def index
+        pagy, keywords_list = pagy(keywords)
+
+        if keywords.any?
+          render json: KeywordSerializer.new(keywords_list, pagy_options(pagy)).serializable_hash, status: :ok
```
junan: Does it render `200` status code without specifying `status: :ok`?

app/controllers/api/v1/keywords_controller.rb
```diff
@@ -0,0 +1,30 @@
...
+        else
+          render_empty
+        end
+      end
+
+      private
+
+      def render_empty
```
junan: How about moving it to `API::V1::ApplicationController`? then all child controllers can use it, maybe the method can accept the message parameter also, so childs controller can pass message

```ruby
def render_empty(message)
    render json: { meta: message), data: [] }, status: :ok
end
```

app/controllers/concerns/api/v1/pagy/jsonapi_concern.rb
```diff
@@ -0,0 +1,38 @@
...
+            next: url_for_page(pagy.next),
+            prev: url_for_page(pagy.prev),
+            last: url_for_page(pagy.last)
+          }
+        end
+
+        def url_for_page(page)
+          return nil unless page
+
+          url_for only_path: false, params: jsonapi_params.merge({ page: page })
```
junan:
```suggestion
          url_for only_path: false, params: jsonapi_params.merge(page: page)
```

app/serializers/keyword_serializer.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+class KeywordSerializer
+  include JSONAPI::Serializer
+
+  attributes :name
```
junan: How about to pass created at attribute also?
```suggestion
  attributes :name, :created_at
```

app/controllers/api/v1/keywords_controller.rb
malparty: Yes! :) I'll remove the xtra code then :) Thanks!

malparty: Done ;-)

app/controllers/api/v1/keywords_controller.rb
malparty: Moved with a meta_message params ;-)

app/controllers/concerns/api/v1/pagy/jsonapi_concern.rb
malparty: Nice! :) Thanks!

app/serializers/keyword_serializer.rb
malparty: Makes sense! (Y)

config/locales/en.yml
```diff
@@ -35,6 +35,11 @@ en:
     token_revoked: 'If your token was valid, it has been revoked'
   keywords:
     could_not_query: 'An error occurs when performing the Google Search, please try again.'
+    empty_list: 'You do not have any keyword yet.'
+  pagy:
+    errors:
+      overflow: 'There are not that many pages here! Max page is '
```
junan: In Rails you can pass the value to the translation file
```suggestion
      overflow: "There are not that many pages here! Max page is %{max_page}"
```

Then you can call it 
```ruby
I18n.t('pagy.errors.overflow', max_page: e.pagy.last)
```

 instead of interpolating 
 ```ruby
"#{I18n.t('pagy.errors.overflow')} #{e.pagy.last}"
```

File: app/serializers/keyword_serializer.rb (hunk 9 above)
Comment (malparty) [reply app/serializers/keyword_serializer.rb]: API documentation also updated accordingly ;-)

File: spec/requests/api/v1/keywords_controller_spec.rb [hunk 11]
```diff
@@ -0,0 +1,94 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+describe API::V1::KeywordsController, type: :request do
+  describe '#index' do
```
junan: For API we, usually add the request type also
```suggestion
  describe 'GET #index' do
```

config/locales/en.yml
malparty: Very useful! :) Thanks!

spec/requests/api/v1/keywords_controller_spec.rb
malparty: Changed :)

junan:
@malparty The PR is looks good to me, just waiting to be resolved this open conversation https://github.com/malparty/google-search-ruby/pull/49/files#r662023067

app/controllers/api/v1/keywords_controller.rb
ankitkalia1195: @junan @malparty I am fine with both, but such a logic is usually handled on client end, where they check size of data returned. What do you think @malparty

app/controllers/api/v1/keywords_controller.rb
```diff
@@ -0,0 +1,39 @@
...
+    class KeywordsController < ApplicationController
+      include ::Pagy::Backend
+      include API::V1::Pagy::JSONAPIConcern
+      include ErrorHandlerConcern
+
+      def index
+        pagy, keywords_list = pagy(keywords)
+
+        if keywords.any?
+          render json: KeywordSerializer.new(keywords_list, pagy_options(pagy)).serializable_hash, status: :ok
```
ankitkalia1195: @malparty Can you also try without calling `. serializable_hash`, maybe it works without it as well.

app/controllers/api/v1/keywords_controller.rb
```diff
@@ -0,0 +1,26 @@
...
+  module V1
+    class KeywordsController < ApplicationController
+      include ::Pagy::Backend
+      include API::V1::Pagy::JSONAPIConcern
+
+      def index
+        pagy, keywords_list = pagy(keywords)
+
+        if keywords.any?
+          render json: KeywordSerializer.new(keywords_list, pagy_options(pagy)).serializable_hash
```
ankitkalia1195: Can we also try to remove `.serializable_hash`

app/controllers/api/v1/keywords_controller.rb
junan: @ankitkalia1195 @malparty I just checked our previous ([Braive](https://github.com/nimblehq/braive-web/blob/development/app/controllers/api/v1/journals_controller.rb#L10), [The1-API](https://github.com/nimblehq/the1-api/blob/development/app/controllers/v2/districts_controller.rb#L6)) projects, we did not handle empty cases and let the client handle it as just Ankit said.

Both ways are fine to me but now I am preferring more not to handle the empty case and let the client handle it as we already followed this pattern in our projects, it will be better to follow the same for consistency.

app/controllers/api/v1/keywords_controller.rb
malparty: Yes @ankitkalia1195, you are right! It's working! 🚀

app/controllers/api/v1/keywords_controller.rb
malparty: After second reading, yes, I think the call for simplicity wins at the end. I will remove this case ;-)

malparty: Okay, done! ✅

junan (approved):

ankitkalia1195 (approved):

PR #52, opened 2021-06-30
[#9] [Backend] As a User, I can upload a csv file in order to run multiple keyword searches

### LINKED ISSUE #9: [Backend] As a User, I can upload a csv file in order to run multiple keyword searches
## Why?

### Additional notes:
Problems to solve:

No Gos:
- No fancy drag & drop
- No multiple files upload
- No search/filters [yet]

## Who Benefits?
Logged Users.

malparty:
- #9 

## What happened

Add `keywords#create` action to handle CSV file upload & db persistence.
- Include file and keywords validations
- Perform a single SQL INSERT while still uses Active Model validation

#### CSV File upload
- `CSVUploadForm` will handle the `params[:file]` in `keywords#create` requests
- This action redirects to `keywords#index` (root path) with either a success message, either a list of errors
- We use Ruby build-in `csv` module to parse csv files
    - 1 row = 1 keyword
    - In case many columns, they are joined into a single keyword (to enable searching google including `,` )

#### Validation
- Done by `CSVValidator`
- Handle csv file with 3 checks:  File type / File Blank / Keyword Count 
- Handle keywords using Keyword Model Validation (and promote keywords errors into itself)

#### Performance
I use 'insert_all' to perform a single INSERT with multiple values
- Validation is done BEFORE `insert_all`
- Keyword validation perform SQL queries to validate the `user` association. This would result in unacceptable performances for 200+ keywords files (2.5ms per keyword validation!). So I added a flag `:bulk_insert` inside keyword model, that would bypass this validation. 

#### Tests cover all paths:

image

#### SQL Query are optimized:

- For 8 keywords - approx 2ms, no other queries
    image

- For 1000 keywords - approx 26ms, no other queries
image


app/models/keyword.rb
```diff
@@ -4,8 +4,17 @@ class Keyword < ApplicationRecord
   include Discard::Model
 
   validates :name, presence: true, length: { maximum: 255 }
+  validates :user, presence: true, unless: :bulk_insert?
```
ankitkalia1195: Why do we need to disable this validation in case of bulk_insert? 🤔

ankitkalia1195: When we run bulk insert, the rails validations are skipped anyways. Only database level constraints matter

malparty: @ankitkalia1195: short answer: in order to use the keyword model validation **before** insert_all `keywords.select(&:validate)` -- this validation perform 1x SQL query per keyword (to verify the user) which destroys the performances of the app (in the context of 200 to 1000 keywords)

Long answer: If we base only on DB level validation, DB will trim any keyword longer than 255 (not to mention that the whole model validation is set for nothing). 
Though, when doing the model validation, for performance reason, I still need to bypass the user association validation (which is well managed by DB -- so it should not be a big deal).

ankitkalia1195: @malparty Got it, altough it can be good to checkout db level constraints for this case. https://blog.saeloun.com/2021/01/08/rails-6-check-constraints-database-migrations.html

```
   add_check_constraint :keywords, "char_length(name) <= 255", "char_length_name"
```

malparty: @ankitkalia1195 wise solution. So to confirm, the idea is to not use keyword model validation at all and simply rescue DB exceptions to show eventual errors to the user, right? 
I'll update the code accordingnly. 
Meanwhile I set this PR back to `WIP`. Thanks!

ankitkalia1195: @malparty Yea, if we are using bulk insert, then we can skip the validations all together. But on our side we should be aware of what validations and callbacks we are skipping

malparty: @ankitkalia1195  thanks, very insightful :) 
I have updated the code accordingly:
- Rescue DB error and add it in the Form (so that it can be displayed later)
- Add tests to ensure the transaction does not save any keyword if one of them could not be saved
- Remove "blank" validation and simply ignore any blank keyword (opportunity to discover `Array.filter_map` :) 

The PR is back to "in review" (no more WIP). 
Thanks :)

app/forms/csv_upload_form.rb
```diff
@@ -0,0 +1,55 @@
...
+    end
+
+    errors.empty?
+  end
+
+  private
+
+  attr_reader :user
+
+  def parse_keywords
```
ankitkalia1195: Can the function name can be changed to `parsed_keywords`, as we are depending on its return value.

malparty: Thanks! I'll be more cautious about naming based on returned value :)

app/forms/csv_upload_form.rb
```diff
@@ -0,0 +1,55 @@
...
+
+  def save(params)
+    @file = params[:csv_upload_form][:file]
+
+    return false unless valid?
+
+    begin
+      Keyword.transaction do
+        # rubocop:disable Rails/SkipsModelValidations
+        user.keywords.insert_all parsed_keywords
```
junan, line 24: I think we need the bang method here
```suggestion
        user.keywords.insert_all! parsed_keywords
```

Without `!` method, it will not raise any exception and rails database translation will not work. What do you think?

app/forms/csv_upload_form.rb
```diff
@@ -0,0 +1,55 @@
+# frozen_string_literal: true
+
+require 'csv'
+
+class CSVUploadForm
+  include ActiveModel::Validations
+
+  attr_reader :file
```
junan, line 8: We are not accessing `file` outside of this class. I think we can move it to private section

app/forms/csv_upload_form.rb
```diff
@@ -0,0 +1,55 @@
...
+  end
+
+  def keyword_attributes(name, time)
+    return nil if name.blank?
+
+    {
+      user_id: user.id,
+      name: name,
+      created_at: time,
+      updated_at: time
```
junan: Rails handle for us setting `created_at`, `updated_at `
```suggestion
```

db/migrate/20210702143800_add_length_constraint_to_name_to_keywords.rb
```diff
@@ -0,0 +1,6 @@
+class AddLengthConstraintToNameToKeywords < ActiveRecord::Migration[6.1]
+  def change
+    add_check_constraint :keywords, 'char_length(name) <= 255', name: 'char_length_name'
+  end
+end
+
```
junan:
```suggestion
```

app/forms/csv_upload_form.rb
malparty: good insight :) 
I've check at the [documentation](https://apidock.com/rails/ActiveRecord/Persistence/ClassMethods/insert_all%21), I believe we do not need the bang here:
The bang only adds this:
> Raises ActiveRecord::RecordNotUnique if any rows violate a unique index on the table. In that case, no rows are inserted.
To skip duplicate rows, see ActiveRecord::Persistence#insert_all. To replace them, see ActiveRecord::Persistence#upsert_all.

In this situation, there is no risk of duplicates: Each ID is incrementally generated by the DB. If a user upload a CSV file with 20 same keywords, I believe he wants to compare how regular google ads works, in which case we will consider these 20 keywords as different.

Regarding exception catching for data validation, as the tests show, the DB is still throwing errors and no single keyword is being created if a file contains any invalid keyword. This was a strong focus point during my tests ;-)

So I think it is good without the bang. What do you think @junan ? :)

app/forms/csv_upload_form.rb
malparty: @junan : Not when using `insert_all`, it would raise an error~ 
I had the same though but trials and fail proved me the opposite ^^

malparty: Though [this discussion](https://github.com/rails/rails/issues/35493) points to an interesting workarround: I can add an explicit default value to the DB columns.
I think it's an interesting move... @junan & @ankitkalia1195 do you see any issue with that solution? :)

app/forms/csv_upload_form.rb
malparty: It is actually used by the validation ([CSVValidator](https://github.com/malparty/google-search-ruby/blob/8c7b2198c036f8d555afecd65a8fb2b9d675eeb0/app/validators/csv_validator.rb#L16)). Ex:
```ruby
  def file
    csv_form.file
  end
```
So I think we need to keep it. 
Let me know @junan, maybe I missed a trick from Rails :) 
Thanks!

app/forms/csv_upload_form.rb
ankitkalia1195: @malparty Yea the solution seems good

malparty: @ankitkalia1195 @junan 
Solution implemented [here](https://github.com/malparty/google-search-ruby/pull/52/commits/6e5d5a5b103dd69b395fbf5f9d80f1813e7799bb). 
For migration, I could not find a ruby-way to handle it properly, so I used SQL execute. I hope it's okay. 

Thanks :)

db/migrate/20210702143800_add_length_constraint_to_name_to_keywords.rb
```diff
@@ -0,0 +1,8 @@
+class AddLengthConstraintToNameToKeywords < ActiveRecord::Migration[6.1]
+  def change
+    add_check_constraint :keywords, 'char_length(name) <= 255', name: 'char_length_name'
+
+    execute("ALTER TABLE keywords ALTER COLUMN created_at SET DEFAULT CURRENT_TIMESTAMP")
```
ankitkalia1195: @malparty I guess this part has nothing to do with `AddLengthConstraintToNameToKeywords`, so better create a seperate migration. Also, it should be reversible migration, so we would need to define both up and down methods in this case.

```
   class SetDefaultValueForKeywordTimestamps < ActiveRecord::Migration[6.1]
      def up
        execute("ALTER TABLE keywords ALTER COLUMN created_at SET DEFAULT CURRENT_TIMESTAMP")
     end

     def down
       ## Reverse this migration
     end
end
```

malparty: Smart remark! Thanks, it's done ;-)

app/forms/csv_upload_form.rb
junan: My actual concern was with rails `transaction` block where(inside the block) we are not raising any exception manually or using `!`(which raises an expectation when it can't create a record rather than just returning false) when creating records.  And the purpose of `transaction` is to create all records or don't create any record if one record is failed to create and for that we usually need to raise an exception when one record is failed to create and it is easier to use bang `!` method to raise an exception rather than manually raise an exception.

I checked the code why it is working even you have not used `!` or raising manually and exception for this case.  I found that `insert_all! / insert_all ` skips validations(model validation), then it raises database null constraints exception if you insert invalid data because we set `null: false` constraints on `name` field in the database. That's why it was working as it raises exceptions at the database level.

Regarding the `transaction` example

This bellow code will not work as it does not raise an exception and end up creating first user.
```ruby
Keyword.transaction do
  [FFaker::Internet.email, ""].each do |email|
    User.create(last_name: FFaker::Name.last_name,
                first_name: FFaker::Name.first_name,
                email: email,
                password: 'password123',
                password_confirmation: 'password123')
  end
end
```

And if you raise an exception using `!`(or manually) then it will work as it will not create any record
```ruby
Keyword.transaction do
  [FFaker::Internet.email, ""].each do |email|
    User.create!(last_name: FFaker::Name.last_name,
                first_name: FFaker::Name.first_name,
                email: email,
                password: 'password123',
                password_confirmation: 'password123')
  end
end
```
As `insert_all / insert_all! ` skips the validation and database raising the exception, I don't see also big benefit of using `!` in this case. So I am fine with both versions (bang or non bang). Its up to you which one you use chose

malparty: yes exactly, we share the same understanding :) 
Thanks for all these details, it confirmed what I read. 
So as we said earlier: if both bang/no-bang are fine, let's prefer the no-bang ;-)

db/migrate/20210707115300_add_default_timestamps_to_keywords.rb
```diff
@@ -0,0 +1,10 @@
+class AddDefaultTimestampsToKeywords < ActiveRecord::Migration[6.1]
+  def up
+    execute("ALTER TABLE keywords ALTER COLUMN created_at SET DEFAULT CURRENT_TIMESTAMP;")
+    execute("ALTER TABLE keywords ALTER COLUMN updated_at SET DEFAULT CURRENT_TIMESTAMP;")
+  end
+  def down
```
ankitkalia1195, line 7: A blank line can be added between these 2 methods.

malparty: Woups!

ankitkalia1195 (approved):

db/seeds.rb
```diff
@@ -16,5 +16,5 @@
 if Rails.env.development?
   user = User.where(email: 'admin@nimblehq.co').first_or_create(Fabricate.attributes_for(:admin, email: 'admin@nimblehq.co'))
 
-  Fabricate.times(100, :keyword, user: user)
+  Fabricate.times(100, :keyword_with_result, user: user)
```
junan: I think we dont have a factory for `:keyword_with_result `

malparty: Well catched! It's a commit made in the wrong branch =/ 
Thanks!!

malparty: Fixed!

junan (approved):

PR #55, opened 2021-07-02
[#13] [Frontend] As a User, I can upload a csv file in order to run multiple keyword searches

### LINKED ISSUE #13: [Frontend] As a User, I can upload a csv file in order to run multiple keyword searches
Build advanced UI for #9 

## Why?

### Additional notes:
Problems to solve:

No Gos:
- No fancy drag & drop
- No multiple files upload
- No search/filters [yet]

## Who Benefits?
Logged Users.

malparty:
- #13 

## What happened

- Add UI for uploading CSV Files.
- Displays error or success message.

## Insight

- Add UI for uploading CSV Files.
    - basic form, no submit button (submit is triggered by JavaScript `javascript/screen/index.js` when a file is selected)
    - POST method `keywords#create`. Will redirect to index by adding a flash error/success based on CSV parsing/validation status (from backend)
- Displays error or success message.
    - `errors` flash message added (as a string array)
    - `success` flash message added (as a string)

## Proof Of Work

Form on home page:
image

Error message:
image

Success message (with loading if just-uploaded keywords):
image

System tests covering all use-cases.
image

### DIFF
--- a/app/javascript/components/CSVUploadForm/index.js
+++ b/app/javascript/components/CSVUploadForm/index.js
@@ -5,7 +5,7 @@ const SELECTORS = {
 class CSVUploadForm {
   constructor(csvUploadForm) {
     this.csvUploadForm = csvUploadForm;
-    this.fileInput = csvUploadForm.querySelector(SELECTORS.fileInput)
+    this.fileInput = csvUploadForm.querySelector(SELECTORS.fileInput);
 
     this._setup();
   }
@@ -17,11 +17,11 @@ class CSVUploadForm {
   }
 
   _registerOnFileChangeEvent() {
-    document.addEventListener('DOMContentLoaded', () => {
-      // Csv File upload form - submit when file selected
-      this.fileInput.onchange = () => {
-        this.csvUploadForm.submit();
-      };
-    });
+    // Submit form when a file is selected
+    this.fileInput.onchange = () => {
+      this.csvUploadForm.submit();
+    };
   }
 }
+
+export default CSVUploadForm;
--- a/app/javascript/components/CSVUploadForm/index.js
+++ b/app/javascript/components/CSVUploadForm/index.js
@@ -1,11 +1,11 @@
 const SELECTORS = {
-  fileInput: '#csv_upload_form_file'
+  fileInput: '.form-csv-upload__input-file'
 };
 
 class CSVUploadForm {
-  constructor(csvUploadForm) {
-    this.csvUploadForm = csvUploadForm;
-    this.fileInput = csvUploadForm.querySelector(SELECTORS.fileInput);
+  constructor(elementRef) {
+    this.csvUploadForm = elementRef;
+    this.fileInput = elementRef.querySelector(SELECTORS.fileInput);
 
     this._setup();
   }
--- a/app/javascript/screens/index.js
+++ b/app/javascript/screens/index.js
@@ -1,0 +1,1 @@
+import './keywords';
--- a/app/javascript/screens/index.js
+++ b/app/javascript/screens/index.js
@@ -1,6 +1,6 @@
-document.addEventListener("DOMContentLoaded", (event) => {
+document.addEventListener('DOMContentLoaded', (event) => {
   // Csv File upload form - submit when file selected
-  document.getElementById('csv_upload_form_').onchange = () => {
-    document.getElementById('new_CSVUploadForm').submit()
-  }
+  document.getElementById('csv_upload_form_file').onchange = () => {
+    document.getElementById('new_CSVUploadForm').submit();
+  };
 });
--- a/app/javascript/screens/index.js
+++ b/app/javascript/screens/index.js
@@ -1,6 +1,0 @@
-document.addEventListener('DOMContentLoaded', (event) => {
-  // Csv File upload form - submit when file selected
-  document.getElementById('csv_upload_form_file').onchange = () => {
-    document.getElementById('new_CSVUploadForm').submit();
-  };
-});
--- a/app/javascript/screens/index.js
+++ b/app/javascript/screens/index.js
@@ -1,0 +1,6 @@
+document.addEventListener("DOMContentLoaded", (event) => {
+  // Csv File upload form - submit when file selected
+  document.getElementById('csv_upload_form_').onchange = () => {
+    document.getElementById('new_CSVUploadForm').submit()
+  }
+});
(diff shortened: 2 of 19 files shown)


spec/support/file_upload_helpers.rb
```diff
@@ -18,4 +18,14 @@ def uploaded_file(file_name)
       ActionDispatch::Http::UploadedFile.new({ tempfile: file, type: type })
     end
   end
+
+  module System
+    def submit_file(name)
+      sign_in Fabricate(:user)
```
ankitkalia1195: Why do we need to `sign_in` user inside this method 🤔 This method should just contain part about attaching the file

malparty: Is there a way to start a test while being signed in already? 
This method is not a UI method, it's a helper that will call Devise building `login_as`:

```ruby
    def sign_in(user)
      login_as user, scope: :user

      user
    end
```

It would indeed simplify things if I can have a default behavior to sign in :) 
Let me know @ankitkalia1195  what you think :)

ankitkalia1195: @malparty Its okay to have login statement inside the test, or you can extract that behaviour outside like this. But in rspec we dont focus on reducing lines of code too much, test should be as readable as possible. 

```
  RSpec.configure do |config|
    config.before(:all, authenticated_user: true) do
       sign_in 
    end
  end
```

```
    it "submits the file", authenticated_user: true do
   end
```

malparty: thanks, very great to have that actually, it helps to focus on what matters in the tests. 
I have updated the code accordingly. The sign_in method is still used directly in cases where I need to test the "user" object. 
Should be all good now :)

app/javascript/screens/index.js
```diff
@@ -0,0 +1,6 @@
+document.addEventListener('DOMContentLoaded', (event) => {
```
junan: Lets follow our [JS structure conventions](https://nimblehq.co/compass/development/code-conventions/javascript/#js-only-applications) 👍 

Under the screen, we can make a folder with screen name(ex: `dashboard` or any other name)

Then we can define the screen like (pseudocode)
```js

import SomeComponent from 'Components/SomeComponent';


const SELECTORS = {
  screen: 'body.dashboard',
   promotingBanner: '.promoting-banner',
   keuwordList:  '.keyword_list'
  .....
};

class DashboardScreen {
  constructor() {
    this.promotingBanner = document.querySelector(SELECTORS.promotingBanner);
   ...

    this._setup();
  }

  // Private

  _setup() {
    // Here we initialize all the JS component we need for that screen and implement the functionality inside the component
    // ex:  new CSVUploadForm(this.csvUploadForm);, I think your CSV uploading functionality should go in this component
  
  }
}

let isDashboardScreen = document.querySelector(SELECTORS.screen) !== null;

if (isDashboardScreen) {
  new DashboardScreen();
}

```

You can check the reference also here https://github.com/nimblehq/braive-web/blob/development/app/assets/javascripts/screens/patient/dashboard.js

app/views/keywords/_form_upload.html.erb
```diff
@@ -0,0 +1,10 @@
+<%= form_for(csv_form, as: CSVUploadForm, url: keywords_path, html: { method: :post, multipart: true, class:'form-csv-upload' }) do |f| %>
+  <div class="row mb-3">
+    <div class="col-12 col-lg-4 col-md-6">
+      <div class="custom-file">
+        <%= f.label :file, t('csv.upload_btn'), class: 'custom-file-label' %>
+        <%= f.file_field :file, class: 'form-control custom-file-input' %>
```
junan:
```suggestion
        <%= f.file_field :file,  required: true, class: 'form-control custom-file-input' %>
```

app/javascript/screens/index.js
malparty: Super interesting! :) I'll dive into that!! Thanks!

malparty: @junan @ankitkalia1195 
Ok, I have updated & review the JavaScript code to match.

I'm not sure 100% about keeping the form ID as generated (`#new_CSVUploadForm`) or if I should rather use classes to identify the form/file_input. What would you advise here? 

Thanks! :)

junan: For selecting DOM elements, we mostly used `class` instead of `id`

app/javascript/components/CSVUploadForm/index.js
```diff
@@ -0,0 +1,27 @@
+const SELECTORS = {
+  fileInput: '#csv_upload_form_file'
```
junan:
```suggestion
  FileInput: '#csv_upload_form_file'
```

Screen Shot 2021-07-14 at 6 13 55 PM

app/javascript/components/CSVUploadForm/index.js
```diff
@@ -0,0 +1,27 @@
...
+};
+
+class CSVUploadForm {
+  constructor(csvUploadForm) {
```
junan: We usually use `elementRef` as the constructor argument 
```suggestion
  constructor(elementRef) {
```
Screen Shot 2021-07-14 at 6 12 22 PM

app/views/keywords/index.html.erb
```diff
@@ -1,3 +1,4 @@
-<div class="container">
+<div class="container screen-keywords">
```
junan: Lets use singular class `screen-keyword`

app/javascript/components/CSVUploadForm/index.js
malparty: I'm not sure to understand: `SELECTORS` is already screaming_snake_case and for it's key the convention is `lowerCamelCase`.
image
So it seems to me that it's all good here.

Let me know @junan if I missed your point ~

app/javascript/components/CSVUploadForm/index.js
malparty: Fixed :)

app/views/keywords/index.html.erb
malparty: Ok, then I've renamed the javascript screen class/file too, to keep consistency ;-)

app/javascript/components/CSVUploadForm/index.js
junan: My bad! your existing already fine

junan (approved):

ankitkalia1195 (approved):

PR #56, opened 2021-07-05
[#8] [Backend] As a User, I can see a list of my previous keyword results

### LINKED ISSUE #8: [Backend] As a User, I can see a list of my previous keyword results
## Why?
In order to compare different keywords & review past results.

### Additional notes:
Problems to solve:
- API expose the feature to enable third app to consume it

No Gos:
- No search feature [yet]
- No filter feature
- No multiple keywords [yet]
- No CSV File upload [yet]

## Who Benefits?
Logged Users.

malparty:
- #8 

## What happened

- Add results structure to Keyword model
- Add tests for data validation 

## Insight

- Add results structure to Keyword model
    - counts added as int4 in `keywords` table
    - URL added on a join table `result_links`. 
    > I've prior relational tables over JSONB storage in order to enable more advanced SQL queries for the search feature later
    - Added `status` enum on Keyword: `[pending: 0, parsed: 1, failed: 2]'
    > It's awesome how enums are well embedded with ActiveModel :3
    - `result_links` also has an enum `link_type` to define if the link is a top_ads, page_ados or non_ads one.
- Add tests for data validation 
    - only covering Models validations (Keyword & Result Link)

## Proof Of Work

- Model tests are passing:
image

- Generated DB structure:
    - `keywords` table
    image

    - `result_links` table: 
    image


app/models/result_link.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+class ResultLink < ApplicationRecord
+  belongs_to :keyword, inverse_of: :result_links
+
+  enum link_type: { ads_top: 0, ads_page: 1, non_ads: 2 }
+
+  validates :url, presence: true, allow_blank: false
+  validates :keyword, presence: true
```
ankitkalia1195: Do we need this? Association are already validated for presence in `belongs_to`?

db/migrate/20210705035105_add_results_to_keywords.rb
```diff
@@ -0,0 +1,27 @@
...
+    add_column :keywords, :total_link_count, :integer
+    add_column :keywords, :html, :text
+
+    add_index :keywords, :status
+    add_index :keywords, :ads_top_count
+    add_index :keywords, :ads_page_count
+    add_index :keywords, :non_ads_result_count
+    add_index :keywords, :total_link_count
+
+    create_table :result_links do |t|
```
ankitkalia1195: Creating seperate table like this is also fine, but maybe you can also try postgres array type or postgres jsonb type to store results in keywords table itself. What do you think 🤔 
```
    add_column :keywords, :result_links, :jsonb

    result_links: {
    "non_ads": ["", ""],
    "all_ads": ["", ""],
     "top_ads": ["", ""],
}

Comment (malparty) [reply db/migrate/20210705035105_add_results_to_keywords.rb]: Thanks for this feedback @ankitkalia1195 ! :) 
For JSON/JSONB, I have researched before hands and it is not recommended:
- several articles points out the limitations of DB side calculations possible, which made me think of the advance query needed in coming user stories
- [Nimble compass](https://nimblehq.co/compass/development/code-conventions/ruby-on-rails/#relational-database) recommend `jsonb` for configurations like data which seems far different as link data will be needed by the search feature

Regarding Array type, I need to research more :) 
I'll come back with what I found. 

Though let me know if you disagree with the jsonb, you might have real experiences that prove the contrary and I'd be happy to hear from them (I've only searched for 15 minutes ~ )

File: app/models/result_link.rb (hunk 1 above)
Comment (malparty) [reply app/models/result_link.rb]: image
Rails is telling me we don't! :P 
Well catched!!!

File: db/migrate/20210705035105_add_results_to_keywords.rb (hunk 2 above)
Comment (ankitkalia1195) [reply db/migrate/20210705035105_add_results_to_keywords.rb]: Sure, jsonb can make the search more complex if you want to search by urls. :). So we can keep seperate table.

Comment (malparty) [reply db/migrate/20210705035105_add_results_to_keywords.rb]: Ok, I will keep separate tables, but I will spend some times to practice with jsonb and arrays, for learning purpose (in a local branch only). 
Thanks a lot!! :)

File: app/models/result_link.rb [hunk 3]
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+class ResultLink < ApplicationRecord
+  belongs_to :keyword, inverse_of: :result_links
+
+  enum link_type: { ads_top: 0, ads_page: 1, non_ads: 2 }
+
+  validates :url, presence: true, allow_blank: false
```
ankitkalia1195: why have we added `allow_blank`: false? 🤔

malparty: My bad, I was confused between validators and options to omit validation. 
If I understand well, `allow_blank` is an option meaning that by default it's set to false (validation occurs blank or not blank) but once set to true, validation will not be triggered if the :url is blank. 
Thanks for pointing that out @ankitkalia1195 !! :)

ankitkalia1195 (approved):

app/queries/keywords_query.rb
```diff
@@ -2,7 +2,7 @@
 
 class KeywordsQuery
   def initialize(user)
-    @keywords = user.keywords
+    @keywords = user.keywords.select(Keyword.column_names - ['html'])
```
junan: Rails 6 introduced `excluding` method, how about to use it? I think it is a bit more redable
```suggestion
    @keywords = user.keywords.select(Keyword.column_names.excluding("name"))
```

db/migrate/20210705035105_add_results_to_keywords.rb
```diff
@@ -0,0 +1,27 @@
+class AddResultsToKeywords < ActiveRecord::Migration[6.1]
```
junan: Based on our conventions the migration file name needs to be more meaning full, currently, it seems we are adding the `results` column to the `users` table.

As we are adding multiple columns, the migration file name should be like `...add_status_ ads_top_count...to_keywords`

And creating table `result_links`, we can move it out another migration

#### Migration convention
Screen Shot 2021-07-09 at 9 28 54 AM

db/schema.rb
```diff
@@ -8,7 +8,7 @@
 # migrations from scratch. Old migrations may fail to apply correctly if those
 # migrations use external dependencies or application code.
 #
-# It's strongly recommended that you check this file into your version control system.
+# It"s strongly recommended that you check this file into your version control system.
```
junan: Should not be it `It's`?

db/schema.rb
```diff
@@ -65,7 +87,7 @@
     t.datetime "created_at", precision: 6, null: false
     t.datetime "updated_at", precision: 6, null: false
     t.string "first_name"
-    t.string "last_name"
+    t.string "last_name"""
```
junan:
```suggestion
    t.string "last_name"
```

spec/queries/keywords_query_spec.rb
```diff
@@ -18,7 +18,7 @@
 
       result = described_class.new(user).call
 
-      expect(result.count).to eq(3)
+      expect(result.to_a.size).to eq(3)
```
junan: Does it work without converting to an array? `result.size`

malparty: Thanks, your comment opened my curiosity about "if the same type has both count/size, then what is the difference?!). 
[This post](https://stackoverflow.com/questions/6083219/activerecord-size-vs-count) let me with an open question: Do you have any preference between `length` and `size`? Length seems more specific to this need, while size more generic?
I'd be happy to hear your advises @junan  & @ankitkalia1195  :) 
Thanks!

malparty: (both length and size are fully working in this example by the way~ )

db/schema.rb
malparty: 👀 I'm feeling ashamed I did not spot that ><
As of now, and as discussed with Olivier, I will add an extra-time on reviewing carefully the PR before passing it as non-WIP. Sorry for these ones =/

spec/queries/keywords_query_spec.rb
junan: From the post
> if you already load all entries, say User.all, then you should use length to avoid another db query

As we loaded all the all entries/records(in `result ` ), it seems `length` will be good as it will avoid db query in this case

db/migrate/20210705035105_add_results_to_keywords.rb
malparty: Splitting create table from adding columns makes sense ✅

For file naming, I'm not sure to see what we gain with the extra-long name (appart that I cannot see anymore the table name from the project navigation bar ^^). 
image

I'd be happy to get your point of view if you had any PROS from long namings :) 

Meanwhile, to save time, I've implemented the change (extra-long name is not a blocker either, more a curiosity question for learning :) )

Thanks a lot @junan  :)

spec/queries/keywords_query_spec.rb
malparty: Agreed. Fixed ;-)

app/queries/keywords_query.rb
malparty: Sweeeeeeet! Thanks!

malparty: (total agree for the clarity to read the code!)

db/schema.rb
```diff
@@ -56,6 +67,17 @@
     t.index ["uid"], name: "index_oauth_applications_on_uid", unique: true
   end
 
+  create_table "result_links", force: :cascade do |t|
```
junan, line 70: As we added new migration here, I think the Schema version should change `ActiveRecord::Schema.define(version: 2021_07_07_115300)` 

Maybe you can rollback the new  migrations, and re-run them or maybe you can set up everything(`rails db:migrate:reset`)

db/migrate/20210705035105_add_results_to_keywords.rb
junan: > For file naming, I'm not sure to see what we gain with the extra-long name (appart that I cannot see anymore the table name from the project navigation bar ^^).

This is true, the current convention does not look great when we add many fields at once. I think this is a good case to create an RFC about it in the compass.

> Meanwhile, to save time, I've implemented the change (extra-long name is not a blocker either, more a curiosity question for learning :) )

and for now, I think we can go with it.

malparty: Cool, I will open the RFC to see what people think :) 
Thanks!! :)

db/schema.rb
malparty: db reset already done. But your insight is right! I need to rename the migrations with today's date so they are performed at the end... and that will update the schema.rb version! 👌

malparty: Fixed! :)

junan (approved):

Issue #57, opened 2021-07-06
[FrontEnd] Review layout with offcanvas


malparty:
# Why?

- Improve UI and prepare for coming features
- Use native JavaScript

# Who benefits?

Developers for ease of codding future FrontEnd tasks.
User for modern UI and improved UX

PR #58, opened 2021-07-06
[#57] [FrontEnd] Add JQuery to project

malparty:
- #57 

## What happened

Setup JQuery lib. to be used in any javascript code in the app.

## Insight

- JQuery is setup as global plugin directly in Webpack config.
- bootstrap.native package is not needed anymore, so it has been removed.
- basic includes to `bootstrap/dist/js` are done instead
- es lint config updated to remove false positive of "undefined $"

## Proof Of Work

Tested manually with a bootstrap event catching the `show.bs.modal` to ensure inclusions are well done.

$ is available within javascript files scope.

image

No visual change.
No logical changes for the current features.


malparty:
Closed Due to wrong target branch.

PR #59, opened 2021-07-06
[#57] [FrontEnd] Add JQuery to project

malparty:
- #57 

## What happened

Setup JQuery lib. to be used in any javascript code in the app.

## Insight

- JQuery is setup as global plugin directly in Webpack config.
- bootstrap.native package is not needed anymore, so it has been removed.
- basic includes to `bootstrap/dist/js` are done instead
- es lint config updated to remove false positive of "undefined $"

## Proof Of Work

Tested manually with a bootstrap event catching the `show.bs.modal` to ensure inclusions are well done.

$ is available within javascript files scope.

image

No visual change.
No logical changes for the current features.


ankitkalia1195:
@malparty Why do we need jquery, recent trends tend to move away from jquery and towards Pure javascript.? What do you think 🤔

malparty:
> @malparty Why do we need jquery, recent trends tend to move away from jquery and towards Pure javascript.? What do you think 🤔

I would answer "because I don't want to waste too much time on the UI as it's not the focus for this exercise" ^^

Now if you believe it is important (aka "lots of things to learn here too"), I'm okay to stay on pure JS with the workarround of bootstrap.native package~ 

Let me know what seems best to you @ankitkalia1195 / @junan  :)

malparty:
Ok... you guys just gave me the **_hunger for discovery_**, I will play a bit with bootstrap.native tonight :3 
See you in another PR @ankitkalia1195 @junan I will close this one. No JQuery for the project~

ankitkalia1195:
@malparty Actually the requirements on IC state that we dont need to create any complex UI, just basic and clean is enough . So IMO  the whole IC can be done without any javascript and jquery, if you dont want to focus on the UI 👍

PR #60, opened 2021-07-06
[#57] [FrontEnd] Review layout with offcanvas 

### LINKED ISSUE #57: [FrontEnd] Review layout with offcanvas

malparty:
- #57 

## What happened

Main layout upgrade to anticipate coming FrontEnd tasks (keywords results, search, ...).

## Insight

- Upgrade to Bootstrap 5.0
- Add Bootstrap-Icons
- Upgrade layout to have space for "filters" (will be used for advanced search & order keyword lists later)

## Proof Of Work

> Tests remains unchanged and are still 100% passing ;-)

| Desktop | Mobile |
|---|---|
| Desktop | Mobile |


package.json
```diff
@@ -10,12 +10,14 @@
...
+    "@popperjs/core": "2.9.2",
     "@rails/actioncable": "^6.0.0",
     "@rails/activestorage": "^6.0.0",
     "@rails/ujs": "^6.0.0",
     "@rails/webpacker": "5.2.2",
-    "bootstrap": "4.5.2",
-    "bootstrap.native": "3.0.13",
+    "bootstrap": "5.0.2",
+    "bootstrap-icons": "1.5.0",
+    "bootstrap.native": "4.0.3",
```
junan, line 20: I think we don't need it. We used to use it to avoid `jQuery` but in `bootstrap` `5` they drop `jQuery` dependencies which is now full vanilla Javascript. What do you think?

app/views/keywords/_form_upload.html.erb
```diff
@@ -1,10 +1,9 @@
...
-    </div>
-  </div>
+<%= form_for(csv_form, as: CSVUploadForm, url: keywords_path,
+             html: { method: :post, multipart: true, class: 'form-csv-upload d-inline' }) do |f| %>
+  <%= f.label :file do %>
+    <span class="btn btn-success">
+      <i class="bi bi-upload"></i>
+    </span>
+  <% end %>
+  <%= f.file_field :file, required: true, style: 'display:none;' %>
```
junan: Lets do external style instead inline style  👍

package.json
malparty: Good remark :) 
The only reason I kept BSN was that I wanted to reduce migration effort (as Bootstrap Native was already configured & is actually 100% build and maintained for Bootstrap 5).

Besides, BNS is not totally dead either: they now expose the advantage of being [faster, lighter and more compact](https://thednp.github.io/bootstrap.native/).

My opinion: for this project, I think we can keep as is. For another project, definitely we can use the JS provided by Bootstrap directly. 

Let me know if that's good to you @junan  :)

app/views/keywords/_form_upload.html.erb
malparty: Replaced with `d-none` class from bootstrap~ ;-)

package.json
junan: Ok, fine for me

junan (approved):

ankitkalia1195 (approved):

PR #61, opened 2021-07-07
[#19] [API] As a User, I can upload a csv file in order to run multiple keyword searches

### LINKED ISSUE #19: [API] As a User, I can upload a csv file in order to run multiple keyword searches
Build API for #9

## Why?

### Additional notes:
Problems to solve:

No Gos:
- No fancy drag & drop
- No multiple files upload
- No search/filters [yet]

## Who Benefits?
Logged Users.

malparty:
- #19 

## What happened

Add API endpoints to upload a CSV File and persist Keywords (if valid).

## Insight

Reusing the CSVUploadForm from backend PR. 
Simply adding a route, an action and handling 2 cases: success/error.

Tests cover a valid and an invalid file. 
> Note that I do not test again all types of errors as these ones are tested in the Forms/ spec already.

## Proof Of Work

- Request tests:
    image

- API responses (also available in the [API Documentation](https://documenter.getpostman.com/view/16135891/TzeTKqCb):
    - Success
    ```json
    {
            "meta": "Your file has been uploaded!"
    }
    ```
    - Errors
    ```json
    {
        "errors": {
            "details": [
                "The csv file should contain between 1 and 1000 keywords"
            ],
            "code": "invalid_file",
            "status": "unprocessable_entity"
        }
    }
    ```

### DIFF
--- a/app/controllers/api/v1/keywords_controller.rb
+++ b/app/controllers/api/v1/keywords_controller.rb
@@ -14,7 +14,7 @@ module API
 
       def create
         if csv_form.save(create_params[:file])
-          render json: success_create
+          render json: create_success_response
 
         else
           render_errors(
@@ -39,7 +39,7 @@ module API
         params.permit(:file)
       end
 
-      def success_create
+      def create_success_response
         {
           meta: I18n.t('csv.upload_success')
         }
--- a/app/controllers/api/v1/keywords_controller.rb
+++ b/app/controllers/api/v1/keywords_controller.rb
@@ -15,6 +15,7 @@ module API
       def create
         if csv_form.save(create_params[:file])
           render json: success_create
+
         else
           render_errors(
             details: csv_form.errors.full_messages,
--- a/app/controllers/api/v1/keywords_controller.rb
+++ b/app/controllers/api/v1/keywords_controller.rb
@@ -15,7 +15,6 @@ module API
       def create
         if csv_form.save(create_params[:file])
           render json: create_success_response
-
         else
           render_errors(
             details: csv_form.errors.full_messages,
--- a/app/controllers/api/v1/keywords_controller.rb
+++ b/app/controllers/api/v1/keywords_controller.rb
@@ -16,7 +16,11 @@ module API
         if csv_form.save(create_params[:file])
           render json: success_create
         else
-          render json: build_error(detail: csv_form.errors.full_messages, code: :invalid_file)
+          render_errors(
+            details: csv_form.errors.full_messages,
+            code: :invalid_file,
+            status: :unprocessable_entity
+          )
         end
       end
 
--- a/app/controllers/api/v1/keywords_controller.rb
+++ b/app/controllers/api/v1/keywords_controller.rb
@@ -12,10 +12,32 @@ module API
         render json: KeywordSerializer.new(keywords_list, pagy_options(pagy))
       end
 
+      def create
+        if csv_form.save(create_params[:file])
+          render json: success_create
+        else
+          render json: build_error(detail: csv_form.errors.full_messages, code: :invalid_file)
+        end
+      end
+
       private
 
       def keywords
         KeywordsQuery.new(current_user).call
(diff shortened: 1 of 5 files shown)


app/controllers/api/v1/keywords_controller.rb
```diff
@@ -12,11 +12,38 @@ def index
...
+
+      def csv_form
+        @csv_form ||= CSVUploadForm.new(current_user)
+      end
+
+      def create_params
+        params.permit(:file)
+      end
+
+      def success_create
```
ankitkalia1195: Since we are depending on return value of the function, should we rename it to `creation_success_response` or `create_success_response` ?

spec/requests/api/v1/keywords_controller_spec.rb
```diff
@@ -83,4 +83,58 @@
       end
     end
   end
+
+  describe 'POST #create' do
+    context 'given an invalid file' do
+      it 'does not save any new keyword' do
+        create_token_header
```
ankitkalia1195: Should we do something similar to `authenticated_user: true`, something like

```
   it `does not save any new keyword`, authenticated_api_user: true do
  end
```

malparty: It won't cover all tests (as some needs to assign keywords to the user). 
But it's indeed a good option.
Note that I had to use `:each` instead of `:all` in here: `config.before(:each, ...`. 
The reason is that this helper edits a context-specific object (`request`) to insert the authorization header. 

Working well and overall cleaner than before :) 

**Additional insight:** While it's possible to go further with a helper like `authenticated_api_user: :quantity_keywords` and generate keywords too, I think it brings too much complexity for the context of a test (considering the benefits). 
So I would not go that far at the moment. 
Let me know if you think differently @ankitkalia1195  :) 

Thanks!

app/controllers/api/v1/keywords_controller.rb
malparty: yes, totally better/clearer!! 👌

ankitkalia1195 (approved):

spec/requests/api/v1/keywords_controller_spec.rb
```diff
@@ -83,4 +81,46 @@
       end
     end
   end
+
+  describe 'POST #create' do
+    context 'given an invalid file' do
```
junan: We test happy path cases first, then add unhappy cases
```suggestion
    context 'given a valid file' do
```

app/controllers/api/v1/keywords_controller.rb
```diff
@@ -12,11 +12,38 @@ def index
...
+      end
+
       private
 
       def keywords
         KeywordsQuery.new(current_user).call
       end
+
+      def csv_form
+        @csv_form ||= CSVUploadForm.new(current_user)
```
junan, line 34: I think memoization will not work here, every time a request comes, it will not have the previous request information(unless it is stored somewhere).  What do you think?
```suggestion
  CSVUploadForm.new(current_user)
```

app/controllers/api/v1/keywords_controller.rb
```diff
@@ -12,11 +12,38 @@ def index
         render json: KeywordSerializer.new(keywords_list, pagy_options(pagy))
       end
 
+      def create
+        if csv_form.save(create_params[:file])
+          render json: create_success_response
+
```
junan:
```suggestion
```

app/controllers/api/v1/keywords_controller.rb
junan: Ahh, I see you are accessing it in `render_errors` 👍

spec/requests/api/v1/keywords_controller_spec.rb
malparty: Yes, makes total sense! Thanks! :)

junan (approved):

PR #63, opened 2021-07-08
#62 [Chore] Setup Turbo.hotwired.dev

### LINKED ISSUE #62: [Chore] Setup Turbo.hotwired.dev
## Why?

- To learn about it (yeah, we are in an IC)
- Provides with our user a more dynamic experience without adding much complexity in the Frontend code

## Who benefits?

**Developer:** less JavaScript & Frontend maintenance
**Users:** Faster and more dynamic app

malparty:
- #62

## Issues encountered

- Bootstrap Native is not well supporting "adding element later", as [this post](https://github.com/thednp/bootstrap.native/wiki/FAQs#does-it-work-with-later-added-elements) indicates.
    Workaround: Instead of opening the offcanvas with `data-bs-toggle`, I add a 'click' event-listener (few lines of JavaScript are then needed) and handle the offcanvas opening with Bootstrap Native API.

- Turbo Frames do not handle History API automatically [yet]. There is [an open PR](https://github.com/hotwired/turbo/pull/167), but not sure if it will get merged in later versions or not.
    ⚠️ This is an important point that could be a reason to NOT use Turbo in a client project (as the team needs to chose between poor UX or adding more JavaScript -- which lose a main benefit from Turbo).

> As this project is for learning purpose, I assume it is OK to move forward with full awareness of these limitations. 

A more complete example is shown in this [POC Pull Request](https://github.com/malparty/google-search-ruby/pull/64) which won't be merged.

## What happened

- Install `turbo-rails` gem (using beta as IC project // I would use a stable version for client projects)
- Use Turbo-Frame for keywords list (**1)
- User Turbo-Frame for Filter canvas (**2)

| **1 | **2 | 
|---|---|
| image | image |

## Insight

- Use Turbo-Frame for keywords list
    - Load only the keyword list when navigating between pages
- User Turbo-Frame for Filter canvas
    - Load only the filters when visible to the user (big screen = once tag loaded // mobile = when user toggle the filter canvas)
    Reduce initial loading time if a user simply wants to upload a new file for example.

## Proof Of Work

The keyword list is loaded with the initial load.
When switching from 1 page to another, it will only query the partial view and reload the `#list_keywords` frame.
Filters are lazy loading, so they will be requested only when we display them. Again, only the needed views are loaded (from `filters#index` -- currently returning a simple keyword count for demo purpose).

Turbo_demo

### DIFF
--- a/app/javascript/packs/application.js
+++ b/app/javascript/packs/application.js
@@ -4,9 +4,10 @@
 // that code so it'll be compiled.
 
 import Rails from '@rails/ujs';
-import '@hotwired/turbo-rails';
 import * as ActiveStorage from '@rails/activestorage';
 import 'channels';
+
+import '@hotwired/turbo-rails';
 
 import 'core-js/stable';
 import 'regenerator-runtime/runtime';
--- a/app/javascript/packs/application.js
+++ b/app/javascript/packs/application.js
@@ -4,7 +4,7 @@
 // that code so it'll be compiled.
 
 import Rails from '@rails/ujs';
-import "@hotwired/turbo-rails"
+import '@hotwired/turbo-rails';
 import * as ActiveStorage from '@rails/activestorage';
 import 'channels';
 
--- a/app/javascript/packs/application.js
+++ b/app/javascript/packs/application.js
@@ -4,6 +4,7 @@
 // that code so it'll be compiled.
 
 import Rails from '@rails/ujs';
+import "@hotwired/turbo-rails"
 import * as ActiveStorage from '@rails/activestorage';
 import 'channels';
 
--- a/config/cable.yml
+++ b/config/cable.yml
@@ -1,6 +1,5 @@
 development:
-  adapter: redis
-  url: redis://localhost:6379/1
+  adapter: async
 
 test:
   adapter: test
--- a/config/cable.yml
+++ b/config/cable.yml
@@ -1,5 +1,6 @@
 development:
-  adapter: async
+  adapter: redis
+  url: redis://localhost:6379/1
 
 test:
   adapter: test
--- a/Gemfile
+++ b/Gemfile
@@ -25,6 +25,7 @@ gem 'doorkeeper' # Awesome OAuth 2 provider for your Rails / Grape app
 # Assets
 gem 'webpacker', '~>5.2.0' # Transpile app-like JavaScript
 gem 'sass-rails' # SASS
+gem 'turbo-rails' # Speed of SPA without any JavaScript
 
 # Logging tools
 gem 'colorize' # Ruby gem for colorizing text using ANSI escape sequences
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -439,6 +439,8 @@ GEM
     thor (1.1.0)
     tilt (2.0.10)
     timecop (0.9.4)
+    turbo-rails (0.5.12)
+      rails (>= 6.0.0)
     tzinfo (2.0.4)
       concurrent-ruby (~> 1.0)
     undercover (0.4.3)
@@ -535,6 +537,7 @@ DEPENDENCIES
   spring-commands-rspec
   spring-watcher-listen (= 2.0.1)
   timecop
+  turbo-rails
   tzinfo-data
   undercover
   vcr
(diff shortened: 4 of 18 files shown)


config/cable.yml
```diff
@@ -1,5 +1,6 @@
 development:
-  adapter: async
+  adapter: redis
+  url: redis://localhost:6379/1
```
olivierobert: Is Redis required for Turbo to work? 🤔

malparty: As said on the [Handbook](https://turbo.hotwired.dev/handbook/streams#integration-with-server-side-frameworks):
> Turbo itself is completely backend-agnostic

So it should not be a requirement. 

Though when performing the command `rails turbo:install` from [turbo-rails](https://github.com/hotwired/turbo-rails) gem, it automatically adds these 2 lines. 

At the moment I'm only using `Turbo Frames` which does not use cable.

What if I simply remove these lines and study that in another PR (when working on #14)? :)

app/javascript/helpers/ajaxRequestHelper.js
```diff
@@ -0,0 +1,73 @@
...
+    } else {
+      this.errorCallback(xhr);
+    }
+  }
+
+  _onSuccess(xhr) {
+    this.result = xhr.response;
+
+    if (this.loadInElementId != null) {
+      document.getElementById(this.elementId).innerHTML = this.result;
```
junan: I am not sure where we are using the `hotwire/turbo` feature in this PR 🤔  . We are writing plain JS to request and appending the HTML response to make it SPA. I don't know much  `hotwire/turbo` but should not it handle this stuff itself without writing JS ourselves? 🤔

app/javascript/helpers/ajaxRequestHelper.js
```diff
@@ -0,0 +1,73 @@
+import {XMLHttpRequest} from 'xmlhttprequest';
+
+export class AjaxRequest {
```
junan: As this class doing network requests
Screen Shot 2021-07-12 at 1 49 46 PM
, it should be placed under `adapters` folder instead of `helpers`

app/javascript/helpers/ajaxRequestHelper.js
```diff
@@ -0,0 +1,73 @@
+import {XMLHttpRequest} from 'xmlhttprequest';
```
junan: Not sure why we are using this package for request, it's quite outdated. I think it would be better to use browsers [fetch API](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API) or modern library like [axios](https://github.com/axios/axios)

app/javascript/packs/application.js
```diff
@@ -4,6 +4,7 @@
 // that code so it'll be compiled.
 
 import Rails from '@rails/ujs';
+import '@hotwired/turbo-rails';
```
junan: As this one is from external packages, how about place it in a separate group?  this groups import packages are build in rails.

app/javascript/helpers/ajaxRequestHelper.js
malparty: @junan you are right, this file has nothing to do here. It seems like the rebase did not bring this file deletion from `chore/layout-sidebar` branch. This is probably because I did a simple rebase instead of a rebase onto? 
I will investigate the logs. Meanwhile, deleting this file. 

FYI: Turbo does not happened in Javascript file, only in the views with 
- `<%= turbo_frame_tag [...] %>`
- some `data-turbo` attributes

app/javascript/packs/application.js
malparty: Why not, changed ;-)

app/javascript/helpers/ajaxRequestHelper.js
malparty: yes, confirmed, the rebase of nested branch was the issue. I'm currently studying more in depth `rebase onto` and logs in order to provide with proper & clean rebases for next time. Sorry once again for that =/

app/javascript/helpers/ajaxRequestHelper.js
malparty: Assuming I would have kept this file, after re-thinking, yes it seems more appropriate as an adapter.
I got a bit confused by the `we usually create one file per API resource`, as this AjaxRequest object is more general and does not have any dependency with a specific resource. So it appeared more generic = more a helper at first hand. 
But it totally matches the `async tasks` part of an adapter, so yeah, seems ok! ;-) 

@junan i'm still resolving this conversation as this file was not supposed to be in here.

app/javascript/helpers/ajaxRequestHelper.js
malparty: Well noted if I need that one day! Thanks!

junan (approved):

ankitkalia1195 (approved):

PR #68, opened 2021-07-13
#38 [Backend] As a user, my keywords are being parsed in background

### LINKED ISSUE #38: [Backend] As a user, my keywords are being parsed in background
# Why?
So that I do not need to wait before getting the file-well uploaded response from the application.
So that I can close my web-browser while parsing continue.

### Additional notes:
This is a direct improvement to #9
Work is only backend, there is no frontend related task.

# Who benefits?
Logged users who just uploaded a CSV file.

malparty:
- #38 

## What happened

- Add result_links structure to `Google::ParserService`
- Add google distribute and search jobs
- Add `Google::SearchKeywordJob` tests
- Add `Google::DistributeSearchJob` tests

## Insight

- DistributeSearchJob
    - Enable to distribute KeywordSearch jobs from a list of `{:id, :name}`
- SearchKeywordJob
    - Run a single Keyword `searching` and `parsing`
    - Handle retry (default settings are wait 3seconds and retry up to 5 times)
    - Update status to failed from the first retry
 
## Proof Of Work

✅ Tests are covering (and passing) the following:
```
Google::SearchKeywordJob
  #perform
    given a 422 too many requests error
      retries the job when an error is raised
      does not save any result_links
      does not set any result count
      sets the keyword status as failed
      does not set the html attribute
    given a valid request
      queues the job
      sets the keyword html attribute
      saves all result_links in the DataBase
      sets the keyword status as parsed
      sets the links counts with the right values

Google::DistributeSearchJob
  #perform
    given 3 valid keywords
      queues a DistributedSearch job
      queues 3 SearchKeyword jobs
    given an empty array
      does not queue the SearchKeyword jobs
      queues a DistributedSearch job

Finished in 1.15 seconds (files took 2.14 seconds to load)
14 examples, 0 failures
```

### DIFF
--- a/app/controllers/keywords_controller.rb
+++ b/app/controllers/keywords_controller.rb
@@ -14,7 +14,7 @@ class KeywordsController < ApplicationController
 
   def create
     if save_csv_file
-      Google::DistributeSearchJob.perform_later(csv_form.results)
+      Google::DistributeSearchJob.perform_later(csv_form.keyword_ids)
 
       flash[:success] = I18n.t('csv.upload_success')
     else
--- a/app/controllers/keywords_controller.rb
+++ b/app/controllers/keywords_controller.rb
@@ -13,7 +13,9 @@ class KeywordsController < ApplicationController
   end
 
   def create
-    if csv_form.save(create_params[:csv_upload_form][:file])
+    if save_csv_file
+      Google::DistributeSearchJob.perform_later(csv_form.results)
+
       flash[:success] = I18n.t('csv.upload_success')
     else
       flash[:errors] = csv_form.errors.full_messages
@@ -30,6 +32,10 @@ class KeywordsController < ApplicationController
 
   def csv_form
     @csv_form ||= CSVUploadForm.new(current_user)
+  end
+
+  def save_csv_file
+    csv_form.save(create_params[:csv_upload_form][:file])
   end
 
   def create_params
--- a/app/forms/csv_upload_form.rb
+++ b/app/forms/csv_upload_form.rb
@@ -21,7 +21,7 @@ class CSVUploadForm
     begin
       Keyword.transaction do
         # rubocop:disable Rails/SkipsModelValidations
-        @keyword_ids = user.keywords.insert_all(parsed_keywords).map { |id| id['id'] }
+        @keyword_ids = user.keywords.insert_all(parsed_keywords).map { |keyword_hash| keyword_hash['id'] }
         # rubocop:enable Rails/SkipsModelValidations
       end
     rescue ActiveRecord::StatementInvalid
--- a/app/forms/csv_upload_form.rb
+++ b/app/forms/csv_upload_form.rb
@@ -5,7 +5,7 @@ require 'csv'
 class CSVUploadForm
   include ActiveModel::Validations
 
-  attr_reader :file, :results
+  attr_reader :file, :keyword_ids
 
   validates_with CSVValidator
 
@@ -21,7 +21,7 @@ class CSVUploadForm
     begin
       Keyword.transaction do
         # rubocop:disable Rails/SkipsModelValidations
-        @results = user.keywords.insert_all(parsed_keywords, returning: %w[id name]).to_a
+        @keyword_ids = user.keywords.insert_all(parsed_keywords).map { |id| id['id'] }
         # rubocop:enable Rails/SkipsModelValidations
       end
     rescue ActiveRecord::StatementInvalid
--- a/app/forms/csv_upload_form.rb
+++ b/app/forms/csv_upload_form.rb
@@ -5,7 +5,7 @@ require 'csv'
 class CSVUploadForm
   include ActiveModel::Validations
 
-  attr_reader :file
+  attr_reader :file, :results
 
   validates_with CSVValidator
 
@@ -21,7 +21,7 @@ class CSVUploadForm
     begin
       Keyword.transaction do
         # rubocop:disable Rails/SkipsModelValidations
-        user.keywords.insert_all parsed_keywords
+        @results = user.keywords.insert_all(parsed_keywords, returning: %w[id name]).to_a
         # rubocop:enable Rails/SkipsModelValidations
       end
     rescue ActiveRecord::StatementInvalid
--- a/app/jobs/google/distribute_search_job.rb
+++ b/app/jobs/google/distribute_search_job.rb
@@ -5,8 +5,8 @@ module Google
     queue_as :default
 
     def perform(keyword_ids)
-      keyword_ids.each do |keyword_id|
-        SearchKeywordJob.perform_later(keyword_id)
+      keyword_ids.each_with_index do |keyword_id, index|
+        SearchKeywordJob.set(wait: index * 3).perform_later(keyword_id)
       end
     end
   end
--- a/app/jobs/google/distribute_search_job.rb
+++ b/app/jobs/google/distribute_search_job.rb
@@ -1,0 +1,13 @@
+# frozen_string_literal: true
+
+module Google
+  class DistributeSearchJob < ApplicationJob
+    queue_as :default
+
+    def perform(keywords)
+      keywords.each do |keyword|
+        SearchKeywordJob.perform_later(keyword)
+      end
+    end
+  end
+end
--- a/app/jobs/google/distribute_search_job.rb
+++ b/app/jobs/google/distribute_search_job.rb
@@ -4,9 +4,9 @@ module Google
   class DistributeSearchJob < ApplicationJob
     queue_as :default
 
-    def perform(keywords)
-      keywords.each do |keyword|
-        SearchKeywordJob.perform_later(keyword)
+    def perform(keyword_ids)
+      keyword_ids.each do |keyword_id|
+        SearchKeywordJob.perform_later(keyword_id)
       end
     end
   end
--- a/app/jobs/google/search_keyword_job.rb
+++ b/app/jobs/google/search_keyword_job.rb
@@ -6,7 +6,7 @@ module Google
   class SearchKeywordJob < ApplicationJob
     queue_as :default
 
-    retry_on ClientServiceError, ArgumentError
+    retry_on ClientServiceError, ArgumentError, wait: 12.seconds
 
     def perform(keyword_id)
       keyword = Keyword.where(id: keyword_id).select(:id, :name, :user_id).take
--- a/app/jobs/google/search_keyword_job.rb
+++ b/app/jobs/google/search_keyword_job.rb
@@ -8,22 +8,26 @@ module Google
 
     retry_on ClientServiceError, ArgumentError
 
-    def perform(keyword)
-      html_result = ClientService.new(keyword: keyword['name']).call
+    def perform(keyword_id)
+      keyword = Keyword.where(id: keyword_id).select(:id, :name, :user_id).take
+
+      return unless keyword
+
+      html_result = ClientService.new(keyword: keyword.name).call
 
       raise ClientServiceError unless html_result
 
-      update_keyword keyword['id'], ParserService.new(html_response: html_result).call
-    rescue ClientServiceError, ArgumentError => e
-      job_failed keyword['id']
+      update_keyword keyword_id, ParserService.new(html_response: html_result).call
+    rescue ActiveRecord::RecordNotFound, ClientServiceError, ArgumentError
+      update_keyword_status keyword_id, :failed
 
-      raise e
+      raise
     end
 
     private
 
-    def job_failed(keyword_id)
-      Keyword.update keyword_id, status: :failed
+    def update_keyword_status(keyword_id, status)
+      Keyword.update keyword_id, status: status
     end
 
     def update_keyword(keyword_id, attributes)
--- a/app/jobs/google/search_keyword_job.rb
+++ b/app/jobs/google/search_keyword_job.rb
@@ -9,7 +9,7 @@ module Google
     retry_on ClientServiceError, ArgumentError, wait: 12.seconds
 
     def perform(keyword_id)
-      keyword = Keyword.where(id: keyword_id).select(:id, :name, :user_id).take
+      keyword = Keyword.find keyword_id
 
       return unless keyword
 
@@ -17,26 +17,26 @@ module Google
 
       raise ClientServiceError unless html_result
 
-      update_keyword keyword_id, ParserService.new(html_response: html_result).call
+      update_keyword keyword, ParserService.new(html_response: html_result).call
     rescue ActiveRecord::RecordNotFound, ClientServiceError, ArgumentError
-      update_keyword_status keyword_id, :failed
+      update_keyword_status keyword, :failed
 
       raise
     end
 
     private
 
-    def update_keyword_status(keyword_id, status)
-      Keyword.update keyword_id, status: status
+    def update_keyword_status(keyword, status)
+      keyword.update! status: status
     end
 
-    def update_keyword(keyword_id, attributes)
+    def update_keyword(keyword, attributes)
       Keyword.transaction do
         # rubocop:disable Rails/SkipsModelValidations
-        Keyword.find(keyword_id).result_links.insert_all attributes[:result_links]
+        keyword.result_links.insert_all attributes[:result_links]
         # rubocop:enable Rails/SkipsModelValidations
 
-        Keyword.update keyword_id, attributes.except(:result_links)
+        keyword.update! attributes.except(:result_links)
       end
     end
   end
--- a/app/jobs/google/search_keyword_job.rb
+++ b/app/jobs/google/search_keyword_job.rb
@@ -6,8 +6,6 @@ module Google
   class SearchKeywordJob < ApplicationJob
     queue_as :default
 
-    retry_on ClientServiceError, ArgumentError, wait: 12.seconds
-
     def perform(keyword_id)
       keyword = Keyword.find keyword_id
 
--- a/app/jobs/google/search_keyword_job.rb
+++ b/app/jobs/google/search_keyword_job.rb
@@ -1,0 +1,39 @@
+# frozen_string_literal: true
+
+module Google
+  class ClientServiceError < StandardError; end
+
+  class SearchKeywordJob < ApplicationJob
+    queue_as :default
+
+    retry_on ClientServiceError, ArgumentError
+
+    def perform(keyword)
+      html_result = ClientService.new(keyword: keyword['name']).call
+
+      raise ClientServiceError unless html_result
+
+      update_keyword keyword['id'], ParserService.new(html_response: html_result).call
+    rescue ClientServiceError, ArgumentError => e
+      job_failed keyword['id']
+
+      raise e
+    end
+
+    private
+
+    def job_failed(keyword_id)
+      Keyword.update keyword_id, status: :failed
+    end
+
+    def update_keyword(keyword_id, attributes)
+      Keyword.transaction do
+        # rubocop:disable Rails/SkipsModelValidations
+        Keyword.find(keyword_id).result_links.insert_all attributes[:result_links]
+        # rubocop:enable Rails/SkipsModelValidations
+
+        Keyword.update keyword_id, attributes.except(:result_links)
+      end
+    end
+  end
+end
--- a/app/services/google/parser_service.rb
+++ b/app/services/google/parser_service.rb
@@ -25,12 +25,14 @@ module Google
       {
         ads_top_count: ads_top_count,
         ads_page_count: ads_page_count,
-        ads_top_urls: ads_top_urls,
-        ads_page_urls: ads_page_urls,
         non_ads_result_count: non_ads_result_count,
-        non_ads_urls: non_ads_urls,
         total_link_count: total_link_count,
-        html: html
+
+        html: html,
+
+        result_links: result_links,
+
+        status: :parsed
       }
     end
 
@@ -47,23 +49,37 @@ module Google
     end
 
     def ads_top_urls
-      document.css("##{AD_CONTAINER_ID} .#{ADWORDS_CLASS}").map { |a_tag| a_tag['href'] }
+      document.css("##{AD_CONTAINER_ID} .#{ADWORDS_CLASS}").filter_map { |a_tag| a_tag['href'].presence }
     end
 
     def ads_page_urls
-      document.css(".#{ADWORDS_CLASS}").map { |a_tag| a_tag['href'] }
+      document.css(".#{ADWORDS_CLASS}").filter_map { |a_tag| a_tag['href'].presence }
     end
 
     def non_ads_result_count
-      document.css(NON_ADS_RESULT_SELECTOR).count
+      document.css(NON_ADS_RESULT_SELECTOR).count { |a_tag| a_tag['href'].presence }
     end
 
     def non_ads_urls
-      document.css(NON_ADS_RESULT_SELECTOR).map { |a_tag| a_tag['href'] }
+      document.css(NON_ADS_RESULT_SELECTOR).filter_map { |a_tag| a_tag['href'].presence }
     end
 
     def total_link_count
       document.css('a').count
+    end
+
+    def result_links
+      results = result_link_map(ads_page_urls, :ads_page)
+      results += result_link_map(non_ads_urls, :non_ads)
+      results += result_link_map(ads_top_urls, :ads_top)
+
+      results
+    end
+
+    def result_link_map(urls, type)
+      urls.map do |url|
+        { url: url, link_type: type }
+      end
     end
   end
 end
--- a/config/application.rb
+++ b/config/application.rb
@@ -36,7 +36,7 @@ module GoogleSearchRuby
     config.active_job.queue_adapter = :sidekiq
 
     # Prefix the queue name of all jobs with Rails ENV
-    config.active_job.queue_name_prefix = Rails.env
+    config.active_job.queue_name_prefix = nil
 
     # Compress the responses to reduce the size of html/json controller responses.
     config.middleware.use Rack::Deflater
--- a/spec/forms/csv_upload_form_spec.rb
+++ b/spec/forms/csv_upload_form_spec.rb
@@ -18,6 +18,12 @@ RSpec.describe CSVUploadForm, type: :form do
 
         expect(Keyword.where(name: 'dien may xanh, the gio di dong').count).to eq(1)
       end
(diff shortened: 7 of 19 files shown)


app/jobs/google/search_keyword_job.rb
```diff
@@ -0,0 +1,39 @@
...
+    def perform(keyword)
+      html_result = ClientService.new(keyword: keyword.name).call
+
+      raise ClientServiceError unless html_result
+
+      update_keyword keyword.id, ParserService.new(html_response: html_result).call
+    rescue ClientServiceError, ArgumentError => e
+      job_failed keyword.id
+
+      raise e
```
malparty: @ankitkalia1195  @junan 
I would love to hear about your advises for this: I want to update status `failed` BUT also to enable the retry. 
Is it acceptable to `rescue` an error and to raise it inside the rescue block? Or do you see another way to perform that?

Many thanks! :)

app/jobs/google/search_keyword_job.rb
```diff
@@ -0,0 +1,39 @@
+# frozen_string_literal: true
+
+module Google
+  class ClientServiceError < StandardError; end
```
malparty, line 4: @ankitkalia1195 @junan 
This error relates to `ClientService` but is used by this job only (`SearchKeywordJob`). 
I think it is the right place to put it here, but I would appreciate any feedback or alternative~ 

Alternatives I though about:
- Bring it inside `ClientService` and update `ClientService#call` to raise an error instead of returning `false`. 
    I'm not happy with this as returning false makes the service cleaner and more usable.
- Calling an explicit `retry` instead of raising an error in the job.
    Jobs seems build to handle errors, there is no explicit `retry` method (to my knowledge) and calling a `perform_now` inside the `perform` method seems a bad workaround -- maybe I am wrong?
    
Thanks for your lights :)

app/controllers/keywords_controller.rb
```diff
@@ -13,7 +13,9 @@ def index
   end
 
   def create
-    if csv_form.save(create_params[:csv_upload_form][:file])
+    if save_csv_file
+      Google::DistributeSearchJob.perform_later(csv_form.results)
```
ankitkalia1195: Should we add a test for controller, that it enques the job

app/jobs/google/distribute_search_job.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+module Google
+  class DistributeSearchJob < ApplicationJob
+    queue_as :default
+
+    def perform(keywords)
```
ankitkalia1195: @malparty Usually we never pass whole objects in active job, the reason being that when job is enqued this part is stored in memory. So we should just pass the id or ids.  Refer to this code as example https://github.com/nimblehq/mia-web/blob/development/app/jobs/payments/campaign_charge_processing_job.rb

junan: Yes, `sidekiq` itself recommends passing `id` instead of passing object, the object might gets updated in the database before it process in the job, then the job will process the outdated object. https://github.com/mperham/sidekiq/wiki/Best-Practices#1-make-your-job-parameters-small-and-simple

malparty: I'm just passing Ids (a hash containing :id, :name & :user_id). Not the full keyword. 
But while working on the notifications, I realized that my tests were not passing a hash (while the real code does). 

So I will have to review this PR -- for that reason I'm setting it back to WIP.

Thanks a lot for your insights @ankitkalia1195  @junan !! :)

ankitkalia1195: @malparty Why do we need to pass the hash as well, just list of `keyword_ids` should be enough?

malparty: @ankitkalia1195  Well, if we pass the id only, each search job will need 1 additional SQL query (`select name from keyword where id XXX`). 
While the insset_all fonction enable to natively return more than the `id` column (with the `RETURNING` keyword), I assume that a int + 2x string less than 255char remains quite small. 
My initial arbitration was mainly around: SQL server charge impacts users directly while scheduled tasks charge does not. 

But after reading @junan  article, I understand that even a named property is not recommended as it contains the state of the object (that might change from the time being queued to being processed). 

> [...] Don't pass symbols, named parameters, [...]

I will update accordingly to have only the `keyword_id` triggering the jobs. ;-)

app/jobs/google/search_keyword_job.rb
ankitkalia1195: @malparty This seems to to be the only way, but maybe you can remove `e` and write just `raise`, it will re-raise last expcetion.

app/jobs/google/search_keyword_job.rb
```diff
@@ -0,0 +1,39 @@
...
+      update_keyword keyword['id'], ParserService.new(html_response: html_result).call
+    rescue ClientServiceError, ArgumentError => e
+      job_failed keyword['id']
+
+      raise e
+    end
+
+    private
+
+    def job_failed(keyword_id)
```
ankitkalia1195: The method name can be changed to `update_keyword_status`.

app/jobs/google/search_keyword_job.rb
```diff
@@ -0,0 +1,39 @@
+# frozen_string_literal: true
+
+module Google
+  class ClientServiceError < StandardError; end
+
+  class SearchKeywordJob < ApplicationJob
+    queue_as :default
+
+    retry_on ClientServiceError, ArgumentError
```
ankitkalia1195: What if some other exception occurs, will the job not be retried?

malparty: I will experiment some scenarios and let you know. I know for sure the exception then bubble up to sidekiq, but I need to know exactly if sidekiq performs any retry by default or not. 
Back to you soon!

malparty:
Background Job call added in [API::V1::KeywordsController](https://github.com/malparty/google-search-ruby/pull/68/files#diff-7d00cc43aaf2e771dd8efea4a52134720dfdca8b73776eca01e7c55fd6408457) now that I've rebased onto this :)

app/jobs/google/search_keyword_job.rb
malparty: Will do! :)

app/jobs/google/distribute_search_job.rb
malparty: @ankitkalia1195  @junan 
Code updated accordingly 👍

Note that I've also added a test in the `CSVUploadForm` to assert the returning keyword_ids array ;-)

app/jobs/google/search_keyword_job.rb
malparty: @ankitkalia1195 ✅ Done. 
Though it was harder than expected: I needed to find the proper way to authenticate `request` tests (for web controller, not api) with Devise helpers. The `sign_in` method exposed by Devise helpers was hidden by my previous helpers~ 

But after all, I believe it's good -- and probably cleaner. 

Cf commit [.#38 Add request tests to assert job being enqueued](https://github.com/malparty/google-search-ruby/pull/68/commits/502cfcdba2ee15511c7e165ee68cdd31214589f0)

app/jobs/google/search_keyword_job.rb
malparty: From the logs (bellow) while testing different scenarios, we can see that it only retries for the managed cases (explicit in the `retry_on` method). 

This is interesting, and I believe we should keep the behavior as is: Some non anticipated exceptions might have a negative impact (on data or availability) and we might not want to retry. 
But retry on a `429 too many requests` is useful. 

Now I'm considering, maybe we should not retry on `ArgumentError` (ex: In case of wrong keyword_id).
What do you think @ankitkalia1195 ? :) 



### Logs for raising a standard error:
```
[ActiveJob] [Google::SearchKeywordJob] [ed9632ea-e57c-4447-b7b8-c4c99fdf461b] Performing Google::SearchKeywordJob (Job ID: ed9632ea-e57c-4447-b7b8-c4c99fdf461b) from Test(default) enqueued at  with arguments: 9
[ActiveJob] [Google::SearchKeywordJob] [ed9632ea-e57c-4447-b7b8-c4c99fdf461b] Error performing Google::SearchKeywordJob (Job ID: ed9632ea-e57c-4447-b7b8-c4c99fdf461b) from Test(default) in 0.15ms: StandardError (BAM!):
```

### Logs for raising a handled error:
```
[ActiveJob] [Google::SearchKeywordJob] [8a332423-c4ba-45c5-ac11-c25ebe5744ca] Error performing Google::SearchKeywordJob (Job ID: 8a332423-c4ba-45c5-ac11-c25ebe5744ca) from Test(default) in 12.33ms: Google::ClientServiceError (Google::ClientServiceError):
[ . . . stack trace ]
[ActiveJob] Enqueued Google::SearchKeywordJob (Job ID: 8a332423-c4ba-45c5-ac11-c25ebe5744ca) to Test(default) at 2021-07-16 10:17:32 UTC with arguments: 9
Retrying Google::SearchKeywordJob in 3 seconds, due to a Google::ClientServiceError.
``

File: app/controllers/api/v1/keywords_controller.rb [hunk 7]
```diff
@@ -14,6 +14,8 @@ def index
 
       def create
         if csv_form.save(create_params[:file])
+          Google::DistributeSearchJob.perform_later(csv_form.keyword_ids)
```
junan: As this PR is for web backend, this change should be in the API task.

app/jobs/google/distribute_search_job.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+module Google
+  class DistributeSearchJob < ApplicationJob
+    queue_as :default
+
+    def perform(keyword_ids)
+      keyword_ids.each do |keyword_id|
```
junan: > You need to find a way to work around the limitations of mass-searching keywords as Google prevents it. Be creative.

I think to handle this requirement we need to add some interval between requests as If we do massive requests at the same time, Google blocks the IP. Adding some interval between  request/jobs works.(`perform_later` start the job immediately)
```ruby
     keyword_ids.each_with_index do |keyword_id, index|
        SearchKeywordJob.set(wait: index + 3).perform_later(keyword_id)
     end
```

Then it will enqueue the jobs immediately, but start the job based on set time, after the first job running, 2nd job will run 3 seconds later after running the first job, the third job will run 3 seconds later after running the 2nd job, and so on. It will prevent Google blocking.

What do you think?

app/jobs/google/search_keyword_job.rb
```diff
@@ -0,0 +1,43 @@
...
+
+    def perform(keyword_id)
+      keyword = Keyword.where(id: keyword_id).select(:id, :name, :user_id).take
```
junan: I think here we can just make it simple and make less query with `find` as per [sidekiq](https://github.com/mperham/sidekiq/wiki/Best-Practices#2-make-your-job-idempotent-and-transactional) example
```ruby
 keyword = Keyword.find(keyword)
 ```
 
 Then you can update it directly with the object like
 
 ```ruby
keyword.update!(status: status)
 ```
 
 Instead of `Keyword.update keyword_id, status: status` which still will use the `find` method to find the record, then update it.
  What do you think?

File: app/jobs/google/search_keyword_job.rb [hunk 10]
```diff
@@ -0,0 +1,43 @@
...
+
+      return unless keyword
+
+      html_result = ClientService.new(keyword: keyword.name).call
+
+      raise ClientServiceError unless html_result
+
+      update_keyword keyword_id, ParserService.new(html_response: html_result).call
+    rescue ActiveRecord::RecordNotFound, ClientServiceError, ArgumentError
```
junan, line 19: Thi this we don't need this `rescue` block. If any exception raises, then sidekiq will `retry` again and `keyword` status will be the same it is if the exception raise, So I think we don't need to update to again to  `failed`. What do you think?

app/controllers/api/v1/keywords_controller.rb
malparty: Roger that, I'll create a new issue, move this change into it with a new PR. 
Well identified @junan ~

app/jobs/google/search_keyword_job.rb
ankitkalia1195: @malparty IMO we can skip using retry_on, and let sidekiq retry for any exception. And there are limits to the number of retires sidekiq can perform, which can be changed as well.

ankitkalia1195: Using the `retry_on` adds too many complications, when to retry, when to not, and would make the error handling more complicated as well.

junan: Ya, seems we can skip `retry_on`. In the sidekiq's best [handling section](https://github.com/mperham/sidekiq/wiki/Error-Handling#best-practices), its says to use sidekiq's built-in error handling and retying mechanism 
> Let Sidekiq catch errors raised by your jobs. Sidekiq's built-in retry mechanism will catch those exceptions and retry the jobs regularly. The error service will notify you of the exception. You fix the bug, deploy the fix and Sidekiq will retry your job successfully.

app/jobs/google/distribute_search_job.rb
malparty: Interesting @junan  :) 

I have performed some more tests, it seems that there are several threads running concurrently (4 on my local, 10 by default according to google?). My tests shown that the set `wait:` time is not relative to previous tasks, but it's an absolute time **based from the enqueued time**. So I would prefer setting `index * 3`, so that each t+1 tasks will have an additional 3 seconds waiting time. 

Also, failed jobs will currently be retried every 5 seconds. Considering 4 simultaneous tasks, it's approx 1.2s of wait between retries. We should expect to wait 'more' when retrying (because 99% of the time, the reason why it would failed will be `too many requests`). What about setting 4 tasks * 3 seconds = 12 seconds for retrial time too? :)
Thanks,

malparty: Update: Even 3 seconds seems not enough.
I'm not sure exactly that is the mechanism that bring them to start all together by bach of 4... (4 = number of threads available). 

I will perform more tests with longer wait time.

```bash
00 2021-07-19T06:53:21.043Z pid=70693 tid=35 class=Google::DistributeSearchJob jid=bd72b4adf67169d7fb6f58d2 elapsed=0.393 INFO: done
01 2021-07-19T06:53:22.466Z pid=70693 tid=5t  class=Google::SearchKeywordJob jid=1 INFO: start
02 2021-07-19T06:53:23.645Z pid=70693 tid=5t  class=Google::SearchKeywordJob jid=1 elapsed=1.179 INFO: done
04 2021-07-19T06:53:25.014Z pid=70693 tid=2kp class=Google::SearchKeywordJob jid=2 INFO: start
05 2021-07-19T06:53:26.157Z pid=70693 tid=2kp class=Google::SearchKeywordJob jid=2 elapsed=1.143 INFO: done
09 2021-07-19T06:53:30.411Z pid=70693 tid=2o1 class=Google::SearchKeywordJob jid=3 INFO: start
09 2021-07-19T06:53:30.417Z pid=70693 tid=35  class=Google::SearchKeywordJob jid=4 INFO: start
10 2021-07-19T06:53:31.312Z pid=70693 tid=35  class=Google::SearchKeywordJob jid=4 elapsed=0.895 INFO: done
10 2021-07-19T06:53:31.427Z pid=70693 tid=2o1 class=Google::SearchKeywordJob jid=3 elapsed=1.016 INFO: done
13 2021-07-19T06:53:34.623Z pid=70693 tid=35  class=Google::SearchKeywordJob jid=5 INFO: start
14 2021-07-19T06:53:35.558Z pid=70693 tid=35  class=Google::SearchKeywordJob jid=5 elapsed=0.935 INFO: done
16 2021-07-19T06:53:37.763Z pid=70693 tid=5t  class=Google::SearchKeywordJob jid=6 INFO: start
16 2021-07-19T06:53:37.770Z pid=70693 tid=2kp class=Google::SearchKeywordJob jid=7 INFO: start
17 2021-07-19T06:53:38.757Z pid=70693 tid=2kp class=Google::SearchKeywordJob jid=7 elapsed=0.987 INFO: done
17 2021-07-19T06:53:38.769Z pid=70693 tid=5t  class=Google::SearchKeywordJob jid=6 elapsed=1.005 INFO: done
20 2021-07-19T06:53:41.459Z pid=70693 tid=35  class=Google::SearchKeywordJob jid=8 INFO: start
20 2021-07-19T06:53:41.465Z pid=70693 tid=2o1 class=Google::SearchKeywordJob jid=9 INFO: start
21 2021-07-19T06:53:42.964Z pid=70693 tid=2o1 class=Google::SearchKeywordJob jid=9 elapsed=1.499 INFO: done
22 2021-07-19T06:53:43.895Z pid=70693 tid=35  class=Google::SearchKeywordJob jid=8 elapsed=2.436 INFO: done
26 2021-07-19T06:53:47.571Z pid=70693 tid=35  class=Google::SearchKeywordJob jid=10 INFO: start
26 2021-07-19T06:53:47.578Z pid=70693 tid=5t  class=Google::SearchKeywordJob jid=11 INFO: start
26 21021-07-19T06:53:47.584Z pid=70693 tid=2kp class=Google::SearchKeywordJob jid=12 INFO: start
26 2021-07-19T06:53:47.590Z pid=70693 tid=2o1 class=Google::SearchKeywordJob jid=13 INFO: start
```

malparty: (not that I carefully flush the queues between 2 tests, to ensure there is no previous job that will interfere)

junan: > So I would prefer setting index * 3, so that each t+1 tasks will have an additional 3 seconds waiting time.
> What about setting 4 tasks * 3 seconds = 12 seconds for retrial time too? :)

Sure, we can set more interval time based on our needs.

malparty: I forgot to mentionned that on the left side of the log, you have the cumulated seconds.
26 seconds total wait for the 13th job is much lower than 13*3seconds = 39seconds. So the wait command is weird... humhum, any idea?

malparty: Okay... more research simply indicates that sidekiq is not a second-grade precision tool ^^
I will then keep those values : 3s between each search (as you suggested) and 12 seconds on retrial). 
Thanks @Junan for this interesting feedback :)

malparty: Done~

app/jobs/google/search_keyword_job.rb
malparty: That just created a new concern :D 
Now as the exception is NOT managed at rails code level, but at sidekiq level, all my tests that ensure (status: failed, html not edited, etc...) are failing due to the exception thrown while the test is running. 
I have 2 workarround, but non are `rubocop-proof`:
- wrap the test statment with begin/rescue -- not OK for Rubocop
- have 2x expect (1 expect to raise the error, 2nd expect to test the data) -- not OK for Rubocop

Do you see any clean alternative to handle this case? 
I believe it remains important to ensure
1. Status is set as failed
2. Results (html, counts, links) data is NOT being updated

Thanks for your lights @ankitkalia1195 & @junan  :)

app/jobs/google/search_keyword_job.rb
malparty: That's what I did first, but then observing the logs, it queries the whole object from the DB (including the html column that "can" contain a lot of data). 

So on one hand, I agree with the "these keywords are supposed to be all brand new, so it won't happened".
But on the other hand, I do not like to assume the external context of a class when thinking performances. This job could be used in different scenario where a 1000 keywords already contain big html content.

What do you think @junan ?

app/jobs/google/search_keyword_job.rb
malparty: I'm not sure to understand... 
The only purpose of this rescue is to update the status from :pending to :fail

Do you mean to just keep keywords "pending" even when they failed? It is possible, that would simplify the code, but my initial idea was to notify the user when a keyword failed.

If we do not have a `:failed` status, then maybe we should simply remove the whole column, no? Simply check if a `count` column is NIL will tell us if pending or parsed. 

Let me know what you think? 
Personally I would prefer keeping the 'failed' status~ But this is just my opinion, I'm open to change ~

junan: I got it, you wanted to keep them in `failed` state. Then its fine to me

app/jobs/google/search_keyword_job.rb
junan: > That's what I did first, but then observing the logs, it queries the whole object from the DB (including the html column that "can" contain a lot of data).

@malparty How/Where are you logging?  you mean in the `development.log`?

malparty: yes @junan, I have a tail -f on the dev logs but SQL queries also show up in the console when running the app ;-)

junan: Regarding logging, I think it will be the same log and fewer queries if we use `find`. Because currently, it is
doing many tasks for finding the record, in the first part it is query and loads record in an array(actually: `Keyword::ActiveRecord_Relation`), then it filters with `select`, then it picks the first record with `take`
```Ruby
keyword = Keyword.where(id: keyword_id).select(:id, :name, :user_id).take
```
Using `find` we can just find the record(note that at this point there will not `html` or other attribute data as we have not got those data yet)
```ruby
keyword = Keyword.find(keyword_id)
```
Then we can reuse this record where needed. Currently, we are searching this same record with `find` again in the `update_keyword` and `update_keyword_status` with our reusing. 

An even logging also does not matter in this case as Staging/Production, Rails will use [` config.log_level = :info`](https://github.com/malparty/google-search-ruby/blob/develop/config/environments/production.rb#L53), and will not log SQL queries, it just for development envoronment. You can check it by setting `config.log_level = :info` in the `development.rb`

In my idea, we can simplify it like bellow code, which is more simple and efficient
```Ruby
    def perform(keyword_id)
      keyword = Keyword.find(keyword_id)

      return unless keyword

      html_result = ClientService.new(keyword: keyword.name).call

      raise ClientServiceError unless html_result

      update_keyword keyword, ParserService.new(html_response: html_result).call
    rescue StandardError
      update_keyword_status keyword, :failed

      raise
    end

    private

    def update_keyword_status(keyword, status)
      keyword.update!(status: status)
    end

    def update_keyword(keyword, attributes)
      Keyword.transaction do
        # rubocop:disable Rails/SkipsModelValidations
        keyword.result_links.insert_all attributes[:result_links]
        # rubocop:enable Rails/SkipsModelValidations

        keyword.update!(attributes.except(:result_links))
      end
    end
``` 

@malparty What do you think?

malparty: @junan 
So while I disagree for the "it is doing many tasks for finding the record" (cf picture, only 1 clean select is generated).
image

You mark a big point by pointing out that `update` is actually loading everything again. 

So I have 2 options:
A. Clean performance option: using `Keyword.where(id: 410).update_all({status: :failed})`
image
(and same for updating the attributes)

B. Clean code option, using `find` and voila. :) 

As you pointed the problem out, I will then choose option B. 
In case if was a real project, we would still have observability tools to monitor and be aware if one day it becomes a problem. 

Will be fix soon ;-)

malparty: Fixed! ;-)

app/jobs/google/search_keyword_job.rb
junan: Regarding the testing, you mean testing the `SearchKeywordJob` directly, right?

If so then may be in the `search_job_keyword_spec.rb` we can do test for the failed case like(Pseudocode)
```Ruby
 expect { SearchKeywordJob.perform_now(id) rescue StandardError }.to change(keyword.reload, :status).to 'failed'
```

Can you try it?

And if the test is from the controller level, in that case, I think we can mock it so that it does not enqueue the job like
```Ruby
  describe 'POST #create' do
    before do
      allow(SearchKeywordJob).to receive(:perform_later).and_return(true)
    end
    
    ....
    ....
  end
```

Alternatively, there is a gem([rspec-sidekiq)](https://github.com/philostler/rspec-sidekiq)) also for testing sidekiq job, but I have not used it.

app/forms/csv_upload_form.rb
```diff
@@ -21,7 +21,7 @@ def save(file)
     begin
       Keyword.transaction do
         # rubocop:disable Rails/SkipsModelValidations
-        user.keywords.insert_all parsed_keywords
+        @keyword_ids = user.keywords.insert_all(parsed_keywords).map { |id| id['id'] }
```
ankitkalia1195:
```
   user.keywords.insert_all(parsed_keywords).map { |keyword| keyword['id'] }
```

malparty: Here I would not recommend using `keyword` as postgres returns by default an `array of PK`. Naming this array `keyword` exposes us to a wrong assumption that it contains attributes others than `id` (like name, created_at, ...). 

What about `hash_id`, `id_row`, `returning_id` or `insert_response` @ankitkalia1195 ? :)

app/jobs/google/search_keyword_job.rb
malparty: The first example is not working: The rescue behave as expected (this part is fine!) but `status` is changed in DB, not in the ruby object, so the real 'change' event is triggered when calling the `keyword.reload` method (which is NOT in the expect block).

> Though your example still indicates me that it is acceptable to rescue in a spec  -- in very specific cases like this one 👌

For perform_later:
- enqueuing is already tested in the first test & works well (exception or not) as the exception won't propagate in the test
- other tests aims to test the job logic, so I call `perform_now` which will NOT queue any job (=> it will run as a direct sync call like a basic PORO)

It is good to discover the 'rspec-sidekiq' gem, it might be handy in the future, but for this current IC, I only need the `have_enqueued_job` which is available from `include ActiveJob::TestHelper` 

### So conclusion: 
I will simply have a `begin / rescue` in these 2x specific tests. I will NOT test the retry (I rely on sidekiq tests) and we should be all good! :) 

Many thanks for this detailed insights @junan  :)

app/forms/csv_upload_form.rb
ankitkalia1195: can be `keyword_hash`? 

```
    user.keywords.insert_all(parsed_keywords).map { |keyword_hash| keyword_hash['id'] }
```

malparty: Let's go for this one! :)

junan (approved):

ankitkalia1195 (approved):

PR #72, opened 2021-07-15
✅ [#14] [Frontend] As a User, I get notified on the current progress of my keywords parsing

### LINKED ISSUE #14: [Frontend] As a User, I get notified on the current progress of my keywords parsing
Build advanced UI for #11 

## Why?
In order to see newly parsed results as soon as possible.

Problems to solve:
- Improve user experience and awareness of **his/her** current parsing status (no page refresh needed!)
- Enable user to see quickly **his/her** newly parsed results

## Who Benefits?
Logged in Users

malparty:
- #14 

## What happened

- Add Client subscription to the Google::SearchProgressChannel (turbo-stream tag)
- Add turbo-frame to reception the notification card
- Add UI for `SearchProgress` notification card
- Add system tests

## Insight

- A presenter (`Google::CardSearchProgressPresenter`) is used to help rendering the partial view
- The `turbo_stream_from` html custom tag enables the client to register to our `Google::SearchProgressChannel`channel. No javascript needed! 🤞
- The Turbo frame `card_search_progress` will get its content updated thanks to the instruction that wrap the WebSocket content (the turbo-stream way)

> The UI could have some improvements: I have added a [nice-to-have issue for that](https://github.com/malparty/google-search-ruby/issues/73). 

### UI:

Parsed and "Failed being retried" are displayed from Web Socket information:

image

progress_bar_fast

### Tests:
image

### DIFF
--- a/app/jobs/google/search_progress_job.rb
+++ b/app/jobs/google/search_progress_job.rb
@@ -7,10 +7,20 @@ module Google
     def perform(user_id)
       user = User.find(user_id)
 
-      SearchProgressChannel.broadcast_replace_to "toast_search_progress_stream:user_#{user.id}",
-                                                 target: 'toast_search_progress',
-                                                 partial: 'keywords/toast_search_progress',
-                                                 locals: { pending_count: user.keywords.pending.count }
+      SearchProgressChannel.broadcast_append_to "card_search_progress_stream:user_#{user.id}",
+                                                target: 'card_search_progress',
+                                                partial: 'keywords/card_search_progress',
+                                                locals: { presenter: presenter(user.keywords) }
+    end
+
+    private
+
+    def presenter(keywords)
+      Google::CardSearchProgressPresenter.new(
+        parsed_count: keywords.parsed.where('updated_at > ?', 12.hours.ago).count,
+        failed_count: keywords.failed.count,
+        pending_count: keywords.pending.count
+      )
     end
   end
 end
--- a/app/views/keywords/_card_search_progress.html.erb
+++ b/app/views/keywords/_card_search_progress.html.erb
@@ -1,0 +1,25 @@
+<div id="cardSearchProgress" class="card-search-progress">
+  <div class="card-body">
+    <div class="d-flex justify-content-between">
+      <strong class="mr-auto"><%= t('keywords.parsing_progress') %></strong>
+      <div class="card-search-progress--keywords-left">
+        <%= "#{presenter.pending_count} #{t('keywords.keywords_left')}" %>
+      </div>
+      <div>
+        <small>
+          <%= presenter.local_time %>
+        </small>
+      </div>
+    </div>
+    <div class="progress">
+      <div class="progress-bar card-search-progress--progress-bar-parsed" role="progressbar" style="width: <%= presenter.percent_parsed %>%"
+           aria-valuenow="<%= presenter.percent_parsed %>" aria-valuemin="0" aria-valuemax="100">
+        <%= "#{presenter.parsed_count} #{t('keywords.parsed')}" %>
+      </div>
+      <div class="progress-bar bg-warning card-search-progress--progress-bar-failed" role="progressbar" style="width: <%= presenter.percent_failed %>%"
+           aria-valuenow="<%= presenter.percent_failed %>" aria-valuemin="0" aria-valuemax="100">
+        <%= "#{presenter.failed_count} #{t('keywords.failed')}" %>
+      </div>
+    </div>
+  </div>
+</div>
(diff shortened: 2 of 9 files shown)


app/jobs/google/search_progress_job.rb
```diff
@@ -9,10 +9,20 @@ def perform(user_id)
...
+                                                target: 'card_search_progress',
+                                                partial: 'keywords/card_search_progress',
+                                                locals: { presenter: presenter(user.keywords) }
+    end
+
+    private
+
+    def presenter(keywords)
+      Google::CardSearchProgressPresenter.new(
+        parsed_count: keywords.parsed.where('updated_at > ?', 12.hours.ago).count,
```
malparty, line 20: @ankitkalia1195 @junan 
Is this too much logic inside a Job? 
I think it's a little bit overkill to have a dedicated query object for this case. Besides, it's not 100% frontend (the percentage is, but not the count) so I did not want to put that query inside the presenter.

Any alternative is welcomed :) 
Thanks!

ankitkalia1195: @malparty Whats the logic behind last 12 hours? 🤔

malparty: The idea was to display a progress bar (percentage that move from 1 value to 100%). So we need to compare how many keywords have been parsed to get a percent. If we compare `keywords.parsed.count`, the percentage will become quickly irrelevant due to parsing history (ex: 1000 keyword parsed, you upload a new file of 10x keywords, you expect 10% progress when a keyword is being parsed, not 0.001% :) )

I find it to be a good workaround to keep things simple, with
- stateless backend
- no need to store a value on the DB (whether it is a fileId or something else)

Any alternative is welcome of course :) Thanks @ankitkalia1195

app/jobs/google/search_progress_job.rb
```diff
@@ -9,10 +9,20 @@ def perform(user_id)
...
+      SearchProgressChannel.broadcast_append_to "card_search_progress_stream:user_#{user.id}",
+                                                target: 'card_search_progress',
+                                                partial: 'keywords/card_search_progress',
+                                                locals: { presenter: presenter(user.keywords) }
+    end
+
+    private
+
+    def presenter(keywords)
+      Google::CardSearchProgressPresenter.new(
```
junan, line 21: How about moving this presenter in the `app/presenters` folder so all presenters stay in the same place. Also It will make `SearchProgressJob` lightweight

app/views/keywords/_card_search_progress.html.erb
```diff
@@ -0,0 +1,25 @@
+<div id="cardSearchProgress" class="card-search-progress">
```
junan, line 1: As we already have a class attribute, do we still need `id="cardSearchProgress"`? If no, I think we can remove it

spec/support/disable_animation.rb
```diff
@@ -25,6 +25,8 @@ def call(env)
     private
 
     def html?
+      return false unless @headers['Content-Type']
```
junan: Why do we need to change here? 🤔

app/views/keywords/_card_search_progress.html.erb
malparty: Yes @junan, Turbo-Stream `update` and `replace` methods needs an identified html element (with an id) so that it knows which element to replace/update. 

We use the replace method in `search_progress_job`, via Turbo Channel helpers:
```ruby
SearchProgressChannel.broadcast_replace_to ...
```

spec/support/disable_animation.rb
malparty: Good remark @junan :) 

I added this, because the websocket connections generate requests with empty headers:
image

So without this line, all system tests navigating to the home page (that has web sockets connections) will return a nill error:
image

I'm not sure though if it's an issue from Rails-template or a issue on my web sockets configuration? 👀

junan: For this filtering may be we can use Ruby's safe navigation operator `&` to do it in one line
```
@headers['Content-Type']&.include?('html')
```

And maybe up we can fix it in the template also as it generates an error if the header is empty

malparty: on it :) Thanks!

junan (approved):

spec/support/disable_animation.rb
malparty: Fixed & I will implement it in rails-template now (bug declared [here](https://github.com/nimblehq/rails-templates/issues/297))

ankitkalia1195 (approved):

PR #76, opened 2021-07-19
[#75] [API] As a user, my keywords are being parsed in background

### LINKED ISSUE #75: [API] As a user, my keywords are being parsed in background
# Why?
So that I do not need to wait before getting the file-well uploaded response from the application.
So that I can close my web-browser while parsing continue.

### Additional notes:
This is a direct improvement to #9
Work is only backend, there is no frontend related task.

# Who benefits?
Logged users who just uploaded a CSV file.

malparty:
- #75 

## What happened

- When a user upload a csv file from API, it will start automatically to search google and parse the keywords

## Insight

- Call background worker to start Keyword searches ASAP when a csv file has been uploaded via API
- Add tests to ensure the `DistributeSearchJob` is well enqueued 

## Proof Of Work

image

### DIFF
--- a/app/controllers/api/v1/keywords_controller.rb
+++ b/app/controllers/api/v1/keywords_controller.rb
@@ -20,6 +20,8 @@ module API
 
       def create
         if csv_form.save(create_params[:file])
+          Google::DistributeSearchJob.perform_later(csv_form.keyword_ids)
+
           render json: create_success_response
         else
           render_errors(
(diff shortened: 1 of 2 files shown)


spec/spec_helper.rb
```diff
@@ -10,6 +10,10 @@
     sign_in Fabricate(:user)
   end
 
+  config.before(:each, authenticated_request_user: true) do
```
junan: We config similar stuff in the above, it is possible to use one config for authentication? 🤔

malparty: It is not exactly the same thing...
For system authentication (and to save time) we authenticate once for all tests using this authentication.
For requests tests, we need to re-authenticate at every single test, so I used `each` instead of `all` (using all makes requests test fail).

image

junan (approved):

ankitkalia1195 (approved):

PR #79, opened 2021-07-20
✅ [#12] [Frontend] As a User, I can see a list of my previous keyword results

### LINKED ISSUE #12: [Frontend] As a User, I can see a list of my previous keyword results
Build advanced UI for #8 

## Why?
In order to compare different keywords & review past results.

### Additional notes:
Problems to solve:
- API expose the feature to enable third app to consume it

No Gos:
- No search feature [yet]
- No filter feature
- No CSV File upload [yet]

## Who Benefits?
Logged Users.

malparty:
- #12 

## What happened

- [x] Add keyword#show view
- [x] Add 404 error view 
- [x] Add tests

## Insight

Again, we use Turbo from [Hotwire](https://hotwired.dev/) to load the `keywords#show` action into a new frame: `frame_keyword_show`.
In order to keep things simple, this frame apens right bellow the keyword-list. Pagination is not on top of the keyword list to make it more clear what belongs to what. 

I have used a presenter to format url, date/time and to provide the right icons. 
- tests are provided in the `spec/presenters/keyword_presenter_spec.rb`
- icon methods and `google url` methods are not tested: the logic being too basic it would makes more test-maintenance work than helping much. We can discuss this if you want :) 

### Tests:
| presenter | system |
|---|---|
| image | image |

### UI: 

image

image

### DIFF
--- a/app/presenters/keyword_presenter.rb
+++ b/app/presenters/keyword_presenter.rb
@@ -1,0 +1,73 @@
+# frozen_string_literal: true
+
+class KeywordPresenter
+  attr_reader :keyword
+
+  def initialize(keyword)
+    @keyword = keyword
+  end
+
+  def result_link_types
+    keyword.result_links.group_by(&:link_type)
+  end
+
+  def created_at_date
+    keyword.created_at.strftime '%F'
+  end
+
+  def created_at_time
+    keyword.created_at.strftime '%T'
+  end
+
+  def url_title(url)
+    result = url.sub(%r{https?://(www.)?}, '').split('/')[0]
+
+    return result if result.present?
+
+    'google.com'
+  end
+
+  def google_link
+    "https://google.com/search?q=#{keyword.name}"
+  end
+
+  def result_link_type_icon(link_type)
+    case link_type
+    when :ads_top
+      'bi bi-file-earmark-arrow-up-fill'
+    when :ads_page
+      'bi bi-file-earmark-break-fill'
+    when :non_ads
+      'bi bi-file-check'
+    else
+      'bi bi-file-earmark-text'
+    end
+  end
+
+  def status_icon
+    case keyword.status.to_sym
+    when :pending
+      'bi bi-hourglass-split text-warning'
+    when :parsed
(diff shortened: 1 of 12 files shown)


app/presenters/keyword_presenter.rb
```diff
@@ -0,0 +1,73 @@
...
+      'bi bi-hourglass-split text-warning'
+    when :parsed
+      'bi bi-check-lg text-success'
+    else
+      'bi bi-bug text-danger'
+    end
+  end
+
+  def ads_top_count
+    keyword.ads_top_count || 0
```
junan, line 59: It can be solved in the database level by specifying the default value to `0` for `ads_top_count `, `ads_page_count `, `non_ads_result_count `, `total_link_count ` columns

spec/presenters/keyword_presenter_spec.rb
```diff
@@ -0,0 +1,119 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe KeywordPresenter, type: :presenter do
+  describe '#result_link_types' do
+    it 'returns a list grouped by link_types' do
+      presenter = described_class.new(Fabricate(:keyword_parsed_with_links))
```
junan, line 8: How about to use non persisted object case in this case?(other test cases also in this file) It will speed up test suite
```suggestion
      presenter = described_class.new(Fabricate.build(:keyword_parsed_with_links))
```

junan (approved):

spec/presenters/keyword_presenter_spec.rb
malparty: Nice, very good to know this one!!
Issue #82 added for later as it is not a blocking point.

Thanks @junan

app/presenters/keyword_presenter.rb
malparty: Yes, good remark! 
I've added the issue #83 to handle it if I still have time ~ 
thanks!

ankitkalia1195 (approved):

PR #80, opened 2021-07-21
✅ [#77] [Backend] As a User, I can see a list of my previous keyword results (actions)

### LINKED ISSUE #77: [Backend] As a User, I can see a list of my previous keyword results (actions)
## Why? 
So that a user can access to search results

- Today the previous #8 tasks created the DB model but missed actions in controller (keywords#show and rawviews#show)

## Who Benefits?
Authenticated User

malparty:
- #77 

## What happened

- [x] Add keywords#show action
- [x] Add preview controller with show action for html previews
- [x] Remove Brakeman false positive from displaying non escaped HTML content
- [x] Add tests

## Insight

- keyword#show includes result_links to ensure eager loading
- preview#show is provided with its view as I consider this view as a backend adapter (like a "format.json call for API"). There is no UI.
- 404 error is DRY (will be reused by any other Record Not Found exception)

### Request tests:
image

### preview#show is rendering:
image

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -8,4 +8,12 @@ class ApplicationController < ActionController::Base
 
   before_action :authenticate_user!
   before_action :update_allowed_parameters, if: :devise_controller?
+
+  rescue_from ActiveRecord::RecordNotFound, with: :render_not_found
+
+  private
+
+  def render_not_found
+    render 'shared/404', status: :not_found
+  end
 end
--- a/app/controllers/preview_controller.rb
+++ b/app/controllers/preview_controller.rb
@@ -1,0 +1,17 @@
+# frozen_string_literal: true
+
+class PreviewController < ApplicationController
+  def index
+    keyword = Keyword.find(index_params[:keyword_id])
+
+    render layout: false, locals: {
+      keyword: keyword
+    }
+  end
+
+  private
+
+  def index_params
+    params.permit(:keyword_id)
+  end
+end
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -7,7 +7,7 @@ Rails.application.routes.draw do
 
   devise_for :users, controllers: { registrations: :registrations }
 
-  resources :keywords, only: [:index, :create]
+  resources :keywords, only: [:index, :create, :show]
 
   resources :filters, only: :index
 
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -7,7 +7,9 @@ Rails.application.routes.draw do
 
   devise_for :users, controllers: { registrations: :registrations }
 
-  resources :keywords, only: [:index, :create, :show]
+  resources :keywords, only: [:index, :create, :show] do
+    resources :preview, only: :index
+  end
 
   resources :filters, only: :index
 
--- a/spec/requests/keywords_controller_spec.rb
+++ b/spec/requests/keywords_controller_spec.rb
@@ -5,7 +5,7 @@ require 'rails_helper'
 describe KeywordsController, type: :request do
   describe 'POST #create' do
     context 'given a valid file' do
-      it 'queues a Google::SearchKeyword job', authenticated_request_user: true  do
+      it 'queues a Google::SearchKeyword job', authenticated_request_user: true do
         expect { post :create, params: file_params('valid.csv') }.to have_enqueued_job(Google::DistributeSearchJob)
       end
     end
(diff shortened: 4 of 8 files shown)


app/controllers/preview_controller.rb
```diff
@@ -0,0 +1,17 @@
+# frozen_string_literal: true
+
+class PreviewController < ApplicationController
+  def show
+    keyword = Keyword.find(show_params[:id])
+
+    render partial: 'show', locals: {
```
junan: We directly don't render the partial from the controller, we render always partial view from the main view file if needed.
So I suggest render the main view file by renaming it `show.html` instead of `_show.html`
```suggestion
    render locals: {
```

app/controllers/preview_controller.rb
```diff
@@ -0,0 +1,17 @@
...
+      keyword: keyword
+    }
+  end
+
+  private
+
+  def show_params
+    params.permit :id
```
junan: Lets keep consistent like we did in `keywords_controller.rb`
```suggestion
    params.permit(:id)
```

config/routes.rb
```diff
@@ -7,10 +7,12 @@
 
   devise_for :users, controllers: { registrations: :registrations }
 
-  resources :keywords, only: [:index, :create]
+  resources :keywords, only: [:index, :create, :show]
 
   resources :filters, only: :index
 
+  resources :preview, only: :show
```
junan: For this kind of related route, In Rails we usually create nested routes like
```Ruby
  resources :keywords, only: [:index, :create, :show] do
    resource :preview, only: :show
  end
```
It will create a route `/keywords/:keyword_id/preview` .
(This comment is for improvement and optional)

spec/requests/keywords_controller_spec.rb
```diff
@@ -16,4 +16,26 @@
       end
     end
   end
+
+  describe 'GET #show' do
+    context 'given a valid id' do
+      it 'returns a 200 OK status code' do
```
junan:
```suggestion
      it 'returns a 200 status code' do
```

spec/requests/preview_controller_spec.rb
```diff
@@ -0,0 +1,27 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+describe PreviewController, type: :request do
+  describe 'GET #show' do
+    context 'given a valid id' do
+      it 'returns a 200 OK status code' do
```
junan:
```suggestion
      it 'returns a 200 status code' do
```

app/controllers/application_controller.rb
```diff
@@ -8,4 +8,12 @@ class ApplicationController < ActionController::Base
 
   before_action :authenticate_user!
   before_action :update_allowed_parameters, if: :devise_controller?
+
+  rescue_from ActiveRecord::RecordNotFound, with: :render_not_found
+
+  private
+
+  def render_not_found
+    render 'shared/404', status: :not_found
```
junan, line 17: We already have `404.html` file in `/public/404.html`

malparty: I used the same text. 
But several issues with this static html:
1. I'm not sure exactly when it will be used (only in production environment?)
2. It does not have layout
3. It does not use locals (so content would not be translated if we add more languages)

For that reason, google redirected me to this solution. 
I'd be happy to hear your solutions @junan, if  there is a way to solve point 2 and 3 with the static 404.html file ~ Thanks

app/controllers/preview_controller.rb
malparty: Interesting. I did not know the `layout: false` option and that is the reason why I called for a partial.
It's now fixed ;-)

config/routes.rb
malparty: This is A-MA-ZING! :) I felt a little bit constraint by the convention to remain inside CRUD (create/index/show/...). But nesting routes really solve this!!! Thanks a lot!! 🤩

malparty: Note that this changed:
1. prevew#show is not the 'index' action (because we do not want to specify the keyword_id a second time)
2. `ïd` param becomes `keyword_id` params (wow, Rails magic!!) :D

app/controllers/application_controller.rb
junan: As you want to use a custom error instead of static one, then this looks fine to me

junan (approved):

ankitkalia1195 (approved):

PR #81, opened 2021-07-21
[#10] [Backend] As a User, I can search across keywords and results

### LINKED ISSUE #10: [Backend] As a User, I can search across keywords and results
## Why?
In order to provide insights/learning to the user, from the data gathered. 
While individual keywords results might solve localized problems, our users needs to understand trends across multiple keywords. 

### Additional notes:
Problems to solve:
- Enable the user quickly find a given keyword and see its results
- Enable the user to perform advanced queries in order to learn insights from the data gathered

No Gos:
- Search queries are not persisted [yet]

## Who Benefits?
Logged Users.

malparty:
- #10 

## What happened

- [x] Add url and keyword patterns in KeywordsQuery
- [x] Add pattern tests in KeywordsQuery query tests
- [x] Add link_type in url filter
- [x] Add URL match count
- [x] Add link_type filter tests
- [x] Add match count tests
- [x] Add tests to cover filter validations

## Insight

- Only 3x filters will solve the specification needs (url_pattern, keyword_pattern & link_types). 
    The KeywordsQuery returns a `url_match_count` in addition to the `keywords` array (on which we can get the length).
    - ✅ How many URLs contain the word “technology” in AdWords.
    > url_pattern + link_types (as checkboxes in UI, array in API)
    - ✅ How many times a specific URL shows up in stored search results. 
    > url_pattern + url_match_count
    - ✅ How many keywords have URLs in stored search results with 2 or more “/” or 1 or more “>”. 
    > url_pattern: '(.*\/.*){2}'
- Tests covers all 3 filters as well as the url_match_count

### Tests are covering the following:

| Query Object | Filter model |
|---|---|
| image | image image |

### DIFF
--- a/app/controllers/api/v1/keywords_controller.rb
+++ b/app/controllers/api/v1/keywords_controller.rb
@@ -7,7 +7,7 @@ module API
       include API::V1::Pagy::JSONAPIConcern
 
       def index
-        pagy, keywords_list = pagy(keywords_query.keywords)
+        pagy, keywords_list = pagy(keywords_query.keywords_filtered)
 
         render json: KeywordSerializer.new(keywords_list, pagy_options(pagy))
       end
@@ -35,7 +35,7 @@ module API
       private
 
       def keywords_query
-        @keywords_query ||= KeywordsQuery.new(current_user).call
+        @keywords_query ||= KeywordsQuery.new(current_user.keywords)
       end
 
       def csv_form
--- a/app/controllers/api/v1/keywords_controller.rb
+++ b/app/controllers/api/v1/keywords_controller.rb
@@ -7,7 +7,7 @@ module API
       include API::V1::Pagy::JSONAPIConcern
 
       def index
-        pagy, keywords_list = pagy(keywords)
+        pagy, keywords_list = pagy(keywords_query.keywords)
 
         render json: KeywordSerializer.new(keywords_list, pagy_options(pagy))
       end
@@ -34,8 +34,8 @@ module API
 
       private
 
-      def keywords
-        KeywordsQuery.new(current_user).call
+      def keywords_query
+        @keywords_query ||= KeywordsQuery.new(current_user).call
       end
 
       def csv_form
--- a/app/controllers/keywords_controller.rb
+++ b/app/controllers/keywords_controller.rb
@@ -4,10 +4,12 @@ class KeywordsController < ApplicationController
   include Pagy::Backend
 
   def index
-    pagy, keywords_list = pagy(keywords)
+    pagy, keywords_list = pagy(keywords_query.keywords)
 
     render locals: {
-      pagy: pagy, keywords: KeywordsCollectionPresenter.new(keywords_list),
+      pagy: pagy,
+      keywords: KeywordsCollectionPresenter.new(keywords_list),
+      url_match_count: keywords_query.url_match_count,
       csv_form: csv_form
     }
   end
@@ -34,8 +36,8 @@ class KeywordsController < ApplicationController
 
   private
 
-  def keywords
-    KeywordsQuery.new(current_user).call(index_params)
+  def keywords_query
+    @keywords_query ||= KeywordsQuery.new(current_user).call(index_params)
   end
 
   def csv_form
--- a/app/controllers/keywords_controller.rb
+++ b/app/controllers/keywords_controller.rb
@@ -35,7 +35,7 @@ class KeywordsController < ApplicationController
   private
 
   def keywords
-    KeywordsQuery.new(current_user).call
+    KeywordsQuery.new(current_user).call(index_params)
   end
 
   def csv_form
@@ -44,6 +44,9 @@ class KeywordsController < ApplicationController
 
   def save_csv_file
     csv_form.save(create_params[:csv_upload_form][:file])
+
+  def index_params
+    params.permit(KeywordsQuery::QUERY_PARAMS)
   end
 
   def create_params
--- a/app/controllers/keywords_controller.rb
+++ b/app/controllers/keywords_controller.rb
@@ -4,7 +4,7 @@ class KeywordsController < ApplicationController
   include Pagy::Backend
 
   def index
-    pagy, keywords_list = pagy(keywords_query.keywords)
+    pagy, keywords_list = pagy(keywords_query.keywords_filtered)
 
     render locals: {
       pagy: pagy,
@@ -37,7 +37,7 @@ class KeywordsController < ApplicationController
   private
 
   def keywords_query
-    @keywords_query ||= KeywordsQuery.new(current_user).call(index_params)
+    @keywords_query ||= KeywordsQuery.new(current_user.keywords, index_params)
   end
 
   def csv_form
--- a/app/controllers/keywords_controller.rb
+++ b/app/controllers/keywords_controller.rb
@@ -46,6 +46,7 @@ class KeywordsController < ApplicationController
 
   def save_csv_file
     csv_form.save(create_params[:csv_upload_form][:file])
+  end
 
   def index_params
     params.permit(KeywordsQuery::QUERY_PARAMS)
--- a/app/controllers/keywords_controller.rb
+++ b/app/controllers/keywords_controller.rb
@@ -10,6 +10,7 @@ class KeywordsController < ApplicationController
       pagy: pagy,
       keywords: KeywordsCollectionPresenter.new(keywords_list),
       url_match_count: keywords_query.url_match_count,
+      filter_error: keywords_query.error,
       csv_form: csv_form
     }
   end
@@ -49,7 +50,7 @@ class KeywordsController < ApplicationController
   end
 
   def index_params
-    params.permit(KeywordsQuery::QUERY_PARAMS)
+    params.permit(Filter::QUERY_PARAMS)
   end
 
   def create_params
--- a/app/models/filter.rb
+++ b/app/models/filter.rb
@@ -1,0 +1,38 @@
+# frozen_string_literal: true
+
+class Filter
+  include ActiveModel::Model
+
+  QUERY_PARAMS = [:url_pattern, :keyword_pattern, { link_types: [] }].freeze
+  PATTERN_REGEXP =
+    %r{\A\^?((\[[\\\-/:.\w]+\])|([\\\-/:.\w]+)|(\(([\\\-/:.\w]+)\|?([\\\-/:.\w]+)\)))[+?*]?(\{\d+(,\d?)?\})?\$?\Z}
+    .freeze
+
+  attr_accessor :keyword_pattern, :url_pattern, :link_types
+
+  validates :keyword_pattern, format: { with: PATTERN_REGEXP }, allow_blank: true
+  validates :url_pattern, format: { with: PATTERN_REGEXP }, allow_blank: true
+  validate :link_types_exist?
+
+  def self.from_params(params)
+    Filter.new keyword_pattern: params[:keyword_pattern],
+               url_pattern: params[:url_pattern],
+               link_types: params[:link_types]&.map(&:to_i)&.uniq
+  end
+
+  def result_link_filter_exist?
+    url_pattern.present? || link_types.present?
+  end
+
+  protected
+
+  def link_types_exist?
+    return if link_types_valid?
+
+    errors.add :link_types, I18n.t('filters.value_error')
+  end
+
+  def link_types_valid?
+    link_types.blank? || link_types.all? { |link_type| link_type >= 0 && ResultLink.link_types.keys[link_type] }
+  end
+end
--- a/app/models/filter.rb
+++ b/app/models/filter.rb
@@ -17,7 +17,7 @@ class Filter
   def self.from_params(params)
     Filter.new keyword_pattern: params[:keyword_pattern],
                url_pattern: params[:url_pattern],
-               link_types: params[:link_types]&.map(&:to_i)&.uniq
+               link_types: params[:link_types]&.select(&:present?)&.map(&:to_i)&.uniq
   end
 
   def result_link_filter_exist?
--- a/app/queries/keywords_query.rb
+++ b/app/queries/keywords_query.rb
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 class KeywordsQuery
-  QUERY_PARAMS = %i[url_pattern keyword_pattern].freeze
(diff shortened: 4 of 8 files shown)


app/queries/keywords_query.rb
```diff
@@ -1,19 +1,58 @@
...
 
-  def call
+  def call(params = {})
+    @keyword_pattern = params[:keyword_pattern]
+    @url_pattern = params[:url_pattern]
+    @link_types = params[:link_types]
+
+    where_name
+
+    where_result_links
```
ankitkalia1195: Should by rename these methods to `filter_by_name` and `filter_by_result_links`

app/queries/keywords_query.rb
```diff
@@ -1,19 +1,58 @@
...
     order_by_name
+
+    @url_match_count = result_links_filtered.count
+
+    self
   end
 
   private
 
-  attr_reader :keywords
```
ankitkalia1195, line 14: Why are we not using attr_reader now?

app/queries/keywords_query.rb
```diff
@@ -1,19 +1,58 @@
...
+
+  def where_result_links
+    @keywords = keywords.where(result_links: result_links_filtered) if filter_on_result_link?
+  end
+
+  def result_links_filtered
+    result_links = ResultLink.where(keyword: user.keywords)
+
+    result_links = result_links.where('url ~* ?', url_pattern) if url_pattern.present?
+
```
ankitkalia1195, line 49: Do we really need line break between every call?

app/queries/keywords_query.rb
```diff
@@ -1,19 +1,58 @@
...
     @keywords = user.keywords.select(Keyword.column_names.excluding('html'))
   end
 
-  def call
+  def call(params = {})
+    @keyword_pattern = params[:keyword_pattern]
+    @url_pattern = params[:url_pattern]
+    @link_types = params[:link_types]
+
+    where_name
```
ankitkalia1195: Overall the query pattern in this file seems a bit weird, can you refer the mia project for some references https://github.com/nimblehq/mia-web/tree/development/engines. IMO, we can go with some standard naming like

```
   def initalize(relation: Keywords)
      @relation = Keyword.select(Keyword.column_names.excluding('html'))
   end

   def call
       scope = relation
       scope = filter_by_name(scope) if keyword_pattern.present?
       scope = filter_by_result_links(scope) if filter_on_result_link?
   end

```

ankitkalia1195: This would improve the overall readability, in current code, the instance variable is changed anywhere which makes it difficult to track

app/queries/keywords_query.rb
malparty: It moved from `private` to `public` attr_reader, as I need to return multiple values: The @keywords relation but also the url_match_count.

app/queries/keywords_query.rb
malparty: Very interesting, thank you @ankitkalia1195 

I'm trying to implement it, but I have 2 concerns:
1. I need to return 2x things, the keyword list as well as the url_match_count. What would you recommend?
    - return `self` as currently and access these 2 values from public attr_readers?
    - return an array of both values (it will make the controller code somehow less clean)
    - split into 2x different query objects, one for each value. To keep it DRY, I might need a module to include shared code
2. The `scope` example solve 1 issue (keep a non filtered relation for sub-queries) but it feel like we moved from OOP to a more procedural approach, is that an issue for the query object pattern?

Thanks a lot! @junan  feel free to share your insights too :)

malparty: @ankitkalia1195 @junan I have just pushed a first tentative. Feel free to comment it :) 
Thanks!

junan: I checked the MIA codebase and I like this [RevenueQuery](https://github.com/nimblehq/mia-web/blob/development/engines/mia_admin/app/queries/mia_admin/revenue_reports/revenue_query.rb) implementation. It implemented a separate public method for each required data which is more readable/testable and validates **Single-responsibility principle** for this case(where we need multiple values).

In the current implementation, we are returning the value for a single method(returning `keywords` and `url_match_count` for `call` ) which violates the single-responsibility principle.

In my opinion, [RevenueQuery](https://github.com/nimblehq/mia-web/blob/development/engines/mia_admin/app/queries/mia_admin/revenue_reports/revenue_query.rb) implementation approach is better in this case.(We could access these 2 values from public attr_readers also but I find the RevenueQuery implementation is more reliable and testable)

@malparty @ankitkalia1195  What do you think?

app/controllers/api/v1/keywords_controller.rb
```diff
@@ -7,7 +7,7 @@ class KeywordsController < ApplicationController
       include API::V1::Pagy::JSONAPIConcern
 
       def index
-        pagy, keywords_list = pagy(keywords)
+        pagy, keywords_list = pagy(keywords_query[:keywords])
```
junan: I think these changes should be in the API task

app/queries/keywords_query.rb
```diff
@@ -1,19 +1,56 @@
 # frozen_string_literal: true
 
 class KeywordsQuery
-  def initialize(user)
-    @keywords = user.keywords.select(Keyword.column_names.excluding('html'))
+  QUERY_PARAMS = [:url_pattern, :keyword_pattern, { link_types: [] }].freeze
```
junan: We define the permitted params in the controller as it is related to the controller. So I think we can move it into  `app/controllers/keywords_controller.rb`

app/queries/keywords_query.rb
malparty: Nice!! I did not find this one ☝️  
I totally validate that this will work fine here, thanks @junan. 

I did not know if it was acceptable to have many public methods instead of the generic call for a query object pattern. If it is OK, then hell yeah! :) 

I'll implement it in a few minutes ;-)

spec/queries/keywords_query_spec.rb
```diff
@@ -3,44 +3,220 @@
...
+        users = Fabricate.times(2, :user)
+        Fabricate.times(10, :keyword, user: users[0])
+        Fabricate.times(15, :keyword, user: users[1])
+
+        result = described_class.new(users[0].keywords).call
+
+        expect(result[:keywords].map(&:user_id)).to all(eq(users[0].id))
+      end
+
+      it 'only counts url matches from the initialized user' do
```
junan: Could it be like
```suggestion
      it 'returns only matched URLs from the initialized user' do
```

spec/queries/keywords_query_spec.rb
```diff
@@ -3,44 +3,220 @@
...
+      end
+    end
+
+    context 'given an expression as keyword_pattern' do
+      it 'returns only keywords matching the given pattern' do
+        user = Fabricate(:user)
+
+        %w[World awesome Pool Zipline Zoo].each { |name| Fabricate(:keyword, user: user, name: name) }
+
+        result = described_class.new(user.keywords, { keyword_pattern: '.*o.*' }).call
```
junan: What if a user submits an invalid `keyword_pattern` pattern? will it raise an exception? or it will return all keywords of the user as default?

spec/queries/keywords_query_spec.rb
```diff
@@ -3,44 +3,220 @@
...
+      keyword = Fabricate(:keyword_parsed)
+
+      Fabricate.times(10, :result_link_with_keyword, keyword: keyword, link_type: %i[ads_top non_ads].sample)
+
+      result = described_class.new(keyword.user.keywords, { link_types: %i[ads_top non_ads] }).call
+
+      expect(result[:url_match_count]).to eq(10)
+    end
+
+    it 'does not counts url with a different link_type' do
```
junan:
```suggestion
    it 'does not count URL with a different link_type' do
```

app/queries/keywords_query.rb
malparty: @ankitkalia1195 @junan code updated :)

app/controllers/api/v1/keywords_controller.rb
malparty: @junan It's not adding a new thing, but handling non regression: If I put this change in the API branch, it means we cannot release the backend branch independently. I think it is not a safe idea to extends a non-working part across PR (unless really no choice). 
Note that url_match_count is not added in here, as this indeed is a new feature and so it shall be done in a separate branch. 
Not sure if I have convinced you @junan, let me know :)

app/queries/keywords_query.rb
malparty: I should say I was confused for this one: by default I wanted to put these params in the controller. But they are tightly coupled with this query object: When working in this query object, I need to know what params are available in the filters. Also, if I want to add a new filter param, I do not want to edit 2x different controllers (API + Backend) which is not DRY but also risky. 
That is the reason why I created this CONST here and permit the params still in the controller (as permitting params is definitely the controller responsibility). 

Maybe I should just rename them `FILTER_PARAMS' ? :) Let me know @junan

spec/queries/keywords_query_spec.rb
malparty: Yep, today it raise an exception! I will fix that with an alert message~ Nice catch @junan  :D

spec/queries/keywords_query_spec.rb
malparty: In this test, we do test the `count` value, not the list of Keywords or URLs. 
So eventually we can put `it returns the matched URLs count for the authenticated user`

spec/queries/keywords_query_spec.rb
malparty: ## Update
After several trials, it is quite tricky to rescue the error:
- Cannot rescue the error inside the query object (as this one should NOT execute/load the keywords before pagination with pagy)
- Rescuing the error inside the controller is definitely not DRY and makes the actions quite heavy

I am currently implementing another solution: a simple REGEX to limitate the usage of the pattern, that will prevent having any invalid POSIX Expression send to our DB. This solution seems best because:
- We prevent the error instead of treating it
- Better for performance
- Better for code (everything can be done inside the query object then)

malparty: ## Update 2 
It is working fine, but I'm now having another question: 
> Should I put the filters params into a separate model and apply validation on that model? 
What do you think @ankitkalia1195 @junan ?

Thanks! :) 

PS: PR back to WIP so that I can add tests for the validation.

app/models/filter.rb
```diff
@@ -0,0 +1,34 @@
...
+
+class Filter
+  include ActiveModel::Model
+
+  QUERY_PARAMS = [:url_pattern, :keyword_pattern, { link_types: [] }].freeze
+  PATTERN_REGEXP =
+    %r{\A\^?((\[[\\\-/:.\w]+\])|([\\\-/:.\w]+)|(\(([\\\-/:.\w]+)\|?([\\\-/:.\w]+)\)))[+?*]?(\{\d+(,\d?)?\})?\$?\Z}
+    .freeze
+
+  attr_accessor :keyword_pattern, :url_pattern, :link_types
```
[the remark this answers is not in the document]
malparty: Needed by ActiveModel::Model for validation.

app/queries/keywords_query.rb
malparty: Update: Following Junan solution, we now use public methods~

app/queries/keywords_query.rb
malparty: Update: Moved into `Filter:: QUERY_PARAMS`

spec/queries/keywords_query_spec.rb
malparty: I finally added another Model (`model/filter.rb`) to implement a clean validation of the filters, outside from the Query Object.

app/controllers/api/v1/keywords_controller.rb
junan: For this case, I am fine as you changed the core class(KeywordsQuery), without updating here, the existing API code will not work. 

Usually, we try to do break things(tasks) separate PRs whenever possible. And regarding releasing each PR independently, I think we don't need it. Just like we will not release only the backend PR changes without front end PR changes and for releasing a big single feature, we will need to work on multiple PRs and then will release when all after all PRs merged of that big single feature.

app/controllers/keywords_controller.rb
```diff
@@ -4,10 +4,13 @@ class KeywordsController < ApplicationController
   include Pagy::Backend
 
   def index
-    pagy, keywords_list = pagy(keywords)
+    pagy, keywords_list = pagy(keywords_query.keywords_filtered)
 
     render locals: {
-      pagy: pagy, keywords: KeywordsCollectionPresenter.new(keywords_list),
+      pagy: pagy,
+      keywords: KeywordsCollectionPresenter.new(keywords_list),
```
junan, line 11: For the presenter class where we assign to a variable, we [rename](https://github.com/nimblehq/mia-web/tree/development/engines/mia_admin/app/controllers/mia_admin) it as `..._collection_presenter` or `presenter` . So how about do it same here?

app/models/filter.rb
```diff
@@ -0,0 +1,38 @@
...
+
+  validates :keyword_pattern, format: { with: PATTERN_REGEXP }, allow_blank: true
+  validates :url_pattern, format: { with: PATTERN_REGEXP }, allow_blank: true
+  validate :link_types_exist?
+
+  def self.from_params(params)
```
junan, line 17: How about adding unit test for this method and `result_link_filter_exist?` method?

app/controllers/keywords_controller.rb
malparty: Ok, it makes sense. 
I will rename it in the FrontEnd branch #15 (PR - #85) as this will involve renaming locals in all views :)

malparty: Resolved in [commit d884147](https://github.com/malparty/google-search-ruby/pull/85/commits/d884147abe2a77119148fbd8eba640c94e87ec71)

spec/models/filter_spec.rb
```diff
@@ -0,0 +1,179 @@
...
+    context 'given valid link_types' do
+      it 'is valid' do
+        filter = described_class.new link_types: [ResultLink.link_types[:ads_top], ResultLink.link_types[:non_ads]]
+
+        expect(filter).to be_valid
+      end
+    end
+
+    context 'given a filter from_params' do
+      it 'is a Filter object' do
```
junan:
```suggestion
      it 'returns an Filter object' do
```

junan (approved):

ankitkalia1195 (approved):

PR #85, opened 2021-07-22
[#15] [Frontend] As a User, I can search across keywords and results

### LINKED ISSUE #15: [Frontend] As a User, I can search across keywords and results
Build advanced UI for #10 

## Why?
In order to provide insights/learning to the user, from the data gathered. 
While individual keywords results might solve localized problems, our users needs to understand trends across multiple keywords. 

### Additional notes:
Problems to solve:
- Enable the user quickly find a given keyword and see its results
- Enable the user to perform advanced queries in order to learn insights from the data gathered

No Gos:
- Search queries are not persisted [yet]

## Who Benefits?
Logged Users.

malparty:
- #15 

## What happened

- [x] Add views for filter form and match counts
- [x] Use locals instead of hardcoded english in views
- [x] Add system tests
- [x] Add tests to handle error handling
- [x] Review code after these changes

## Insight

- Filter view:
    - update existing containers set in #57 
    - the form is set to submit via `turbo` (hotwire) and to refresh the `frame_list_keywords` turbo frame:
    `data-turbo="true" data-turbo-frame="frame_list_keywords"`
- match counts view:
    - simply add 2x counts: `url_match_count` (from #10 ) and `pagy.count` (containing the total number keywords across all pages)
- System tests covers the following
    - form & counts presence
    - basic queries
    - no result when querying

### System Tests

image

### Overview

- on the left: Filter panel with keyword_pattern, url_pattern and link_type check_boxes
- on the top-right corner: matching counts for keywords & result-links
image

### Detailed:

| Empty filter form | Filled in form |
|---|---|
| image image | image image |

### Pagination embed filters:
image

> to my surprise, pagy handles current GET requests params and includes them in the pagination links, making the code clean and light :) 
image

### DIFF
--- a/app/views/filters/index.html.erb
+++ b/app/views/filters/index.html.erb
@@ -32,7 +32,7 @@
     <i class="bi bi-info-square"></i>
     <%= t('filters.cheatsheet') %>:
     <ul class="list-group">
-      <li class="list-group-item list-group-item-light"><code>a|b</code> 'a' or 'b'</li>
+      <li class="list-group-item list-group-item-light"><code>(a|b)</code> 'a' or 'b'</li>
       <li class="list-group-item list-group-item-light"><code>a*</code> '', 'a', 'aaaa', ...</li>
       <li class="list-group-item list-group-item-light"><code>a+</code> 'a', 'aaaa', ...</li>
       <li class="list-group-item list-group-item-light"><code>a?</code> 'a' or ''</li>
--- a/app/views/filters/index.html.erb
+++ b/app/views/filters/index.html.erb
@@ -2,7 +2,7 @@
   <p>
     <%= t('filters.total_of_keywords', keywords_count: keywords_count) %>
   </p>
-  <%= form_with url: keywords_path, method: :get, class: 'form-filters', data:{turbo:true, turbo_frame:'frame_list_keywords'} do |form| %>
+  <%= form_with url: keywords_path, method: :get, class: 'form-filters', data: { turbo: true, turbo_frame: 'frame_list_keywords' } do |form| %>
     <div>
       <%= form.label :keyword_pattern, class: 'form-label' do %>
         <i class="bi bi-list-check"></i> <%= t('filters.by_keyword') %>...
--- a/app/views/filters/index.html.erb
+++ b/app/views/filters/index.html.erb
@@ -2,7 +2,7 @@
   <p>
     <%= t('filters.total_of_keywords', keywords_count: keywords_count) %>
   </p>
-  <%= form_with url: keywords_path, method: :get, class: 'form-filters', data: { turbo: true, turbo_frame: 'frame_list_keywords' } do |form| %>
+  <%= form_with url: keywords_path, method: :get, class: 'form-filter', data: { turbo: true, turbo_frame: 'frame_list_keywords' } do |form| %>
     <div>
       <%= form.label :keyword_pattern, class: 'form-label' do %>
         <i class="bi bi-list-check"></i> <%= t('filters.by_keyword') %>...
--- a/app/views/filters/index.html.erb
+++ b/app/views/filters/index.html.erb
@@ -2,34 +2,32 @@
   <p>
     <%= t('filters.total_of_keywords', keywords_count: keywords_count) %>
   </p>
-  <form action="<%= keywords_path %>" method="GET" class="form-filters" data-turbo="true" data-turbo-frame="frame_list_keywords">
+  <%= form_with url: keywords_path, method: :get, class: 'form-filters', data:{turbo:true, turbo_frame:'frame_list_keywords'} do |form| %>
     <div>
-      <label for="keyword_pattern" class="form-label">
+      <%= form.label :keyword_pattern, class: 'form-label' do %>
         <i class="bi bi-list-check"></i> <%= t('filters.by_keyword') %>...
-      </label>
-      <input type="text" name="keyword_pattern" placeholder="type a keyword pattern" class="form-control"/>
+      <% end %>
+      <%= form.text_field :keyword_pattern, placeholder: 'type a keyword pattern', class: 'form-control' %>
     </div>
     <div>
-      <label for="url_pattern" class="form-label mt-3">
-        <i class="bi bi-link-45deg"></i> <%= t('filters.result_link') %>...
-      </label>
-      <input type="text" name="url_pattern" placeholder="type a url pattern" class="form-control"/>
+      <%= form.label :url_pattern, class: 'form-label mt-3' do %>
+        <i class="bi bi-link-45deg"></i>
+        <%= t('filters.result_link') %>...
+      <% end %>
+      <%= form.text_field :url_pattern, placeholder: 'type a url pattern', class: 'form-control' %>
     </div>
     <div class="mt-3">
-      <% ResultLink.link_types.each do |link_type| %>
-        <label>
-          <input type="checkbox" name="link_types[]" id='linkType<%= link_type[0].camelize %>' value="<%= link_type[0] %>">
-          <%= t("link_types.#{link_type[0]}") %>
-        </label>
+      <%= form.collection_check_boxes :link_types, ResultLink.link_types, :last, :first do |check_box| %>
+        <%= check_box.check_box %>
+        <%= check_box.label do %>
+          <%= t("link_types.#{check_box.text}") %>
+        <% end %>
       <% end %>
     </div>
     <div class="d-flex justify-content-end">
-      <button type="submit" class="btn btn-outline-primary mt-3 ">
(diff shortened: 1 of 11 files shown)


app/views/filters/index.html.erb
```diff
@@ -1,3 +1,48 @@
...
-  <p>You have a total of <%= keywords_count %> keywords.</p>
+  <p>
+    <%= t('filters.total_of_keywords', keywords_count: keywords_count) %>
+  </p>
+  <form action="<%= keywords_path %>" method="GET" class="form-filters" data-turbo="true" data-turbo-frame="frame_list_keywords">
+    <div>
+      <label for="keyword_pattern" class="form-label">
+        <i class="bi bi-list-check"></i> <%= t('filters.by_keyword') %>...
+      </label>
+      <input type="text" name="keyword_pattern" placeholder="type a keyword pattern" class="form-control"/>
```
ankitkalia1195: How will we populate the value if user visits directly by url? ```localhost:3000/keywords?keyword_pattern=pattern```

app/views/filters/index.html.erb
```diff
@@ -1,3 +1,48 @@
...
+    </div>
+    <div>
+      <label for="url_pattern" class="form-label mt-3">
+        <i class="bi bi-link-45deg"></i> <%= t('filters.result_link') %>...
+      </label>
+      <input type="text" name="url_pattern" placeholder="type a url pattern" class="form-control"/>
```
ankitkalia1195: @malparty Can we use rails form builders in this form?

app/views/filters/index.html.erb
malparty: @ankitkalia1195, it is possible to update the code and pass the params into the filter Turbo Frame `src` attribute, similar to this:
image
(this partial view is built from keywords#index action, so it is easy to provide it with the local `filter_params`)

I did not implement it at the moment as it is not part of the specifications to save a filtered URL but I will probably make it as a nice-to-have to try & confirm if it works or not ;-)

malparty: Issue added here:
- #87 
- #88

app/views/keywords/_list_keywords.html.erb
```diff
@@ -1,12 +1,28 @@
 <section class="list-keyword" data-turbo="true">
   <%= turbo_frame_tag 'frame_list_keywords' do %>
+    <% if filter_error.blank? %>
+      <div class="d-flex justify-content-end match-counts">
```
junan: Based on our CSS conventions, lets use singular class name(same for other plural css classes)
```suggestion
      <div class="d-flex justify-content-end match-count">
```

app/views/filters/index.html.erb
```diff
@@ -1,3 +1,48 @@
 <%= turbo_frame_tag 'frame_filters' do %>
-  <p>You have a total of <%= keywords_count %> keywords.</p>
+  <p>
+    <%= t('filters.total_of_keywords', keywords_count: keywords_count) %>
+  </p>
+  <%= form_with url: keywords_path, method: :get, class: 'form-filters', data:{turbo:true, turbo_frame:'frame_list_keywords'} do |form| %>
```
junan:
```suggestion
  <%= form_with url: keywords_path, method: :get, class: 'form-filters', data: { turbo: true, turbo_frame: 'frame_list_keywords'} do |form| %>
```

app/views/filters/index.html.erb
```diff
@@ -1,3 +1,48 @@
 <%= turbo_frame_tag 'frame_filters' do %>
-  <p>You have a total of <%= keywords_count %> keywords.</p>
+  <p>
+    <%= t('filters.total_of_keywords', keywords_count: keywords_count) %>
+  </p>
+  <%= form_with url: keywords_path, method: :get, class: 'form-filters', data: { turbo: true, turbo_frame: 'frame_list_keywords' } do |form| %>
```
junan:
```suggestion
  <%= form_with url: keywords_path, method: :get, class: 'form-filter', data: { turbo: true, turbo_frame: 'frame_list_keywords' } do |form| %>
```

app/views/keywords/_list_keywords.html.erb
```diff
@@ -1,12 +1,28 @@
 <section class="list-keyword" data-turbo="true">
   <%= turbo_frame_tag 'frame_list_keywords' do %>
-    <% if keywords.groups.any? %>
+    <% if filter_error.blank? %>
+      <div class="d-flex justify-content-end match-count">
+        <div class="text-muted">
+          <div class="text-end match-count--urls">
```
junan: Should we use singular class here too? ` match-count--url`

junan:
@malparty We need to fix the system spec, its failing now

app/views/filters/index.html.erb
malparty: OMG >< I will add a pre-commit check with `class=("|')[ \w]+s( [ \w])?("|')` in the view director if this continue! 
Sorry for the repetitive omit =/

malparty: Haha working pretty well actually!
image

app/views/filters/index.html.erb
```diff
@@ -1,3 +1,48 @@
 <%= turbo_frame_tag 'frame_filters' do %>
-  <p>You have a total of <%= keywords_count %> keywords.</p>
+  <p>
+    <%= t('filters.total_of_keywords', keywords_count: keywords_count) %>
+  </p>
+  <%= form_with url: keywords_path, method: :get, class: 'form-filter', data: { turbo: true, turbo_frame: 'frame_list_keywords' } do |form| %>
```
ankitkalia1195, line 5: @malparty How about using something like `form_for filter`, where filter is using the model we created in backend PR. Going this way, the values will be prefilled in the form.

malparty: Yes, it would make sense (even if today we do not fill any value). 
I will update that once I will understand the root cause for system test failure ;-)
Thanks @ankitkalia1195  :)

junan: @ankitkalia1195 @malparty Actually `form_with` is a new thing that works both(model and non model), its the combination of `form_for` and `form_tag`.  And `form_for` and `form_tag` both will be removed in the future in favor of `form_with`.

So if needed we can pass the model with `form_with` like

```erb
<%= form_with model: your_object do |form| %>
...
...
<% end %>
```

More detail https://m.patrikonrails.com/rails-5-1s-form-with-vs-old-form-helpers-3a5f72a8c78a, https://stackoverflow.com/a/58112003

malparty: After re-reading this, as I'm already using the form_with tag, I think I will implement de model change only in the nice-to-haves tasks:
- #87 
- #88 

Let me know if that is ok or not @ankitkalia1195  @junan  :) 
Thanks,

ankitkalia1195: Sure, its fine for now @malparty

malparty:
@ankitkalia1195 @junan  conflicts resolved 👍

app/views/filters/index.html.erb
```diff
@@ -1,3 +1,48 @@
...
+    </div>
+    <div class="d-flex justify-content-end">
+      <%= form.button t('filters.submit_btn'), class: 'btn btn-outline-primary mt-3' %>
+    </div>
+  <% end %>
+  <div class="form-text mt-3">
+    <i class="bi bi-info-square"></i>
+    <%= t('filters.cheatsheet') %>:
+    <ul class="list-group">
+      <li class="list-group-item list-group-item-light"><code>(a|b)</code> 'a' or 'b'</li>
```
junan, line 35: `light` is a BEM modifier here, so it should be `--light` instead `-light`
```suggestion
      <li class="list-group-item list-group-item--light"><code>(a|b)</code> 'a' or 'b'</li>
```

Screen Shot 2021-08-03 at 11 24 13 AM

junan (approved):

app/views/filters/index.html.erb
malparty: Hehe, yes well spotted @junan -- it catched my eyes too~
But Bootstrap does not want to follow our conventions :D
https://getbootstrap.com/docs/5.0/components/list-group/#contextual-classes

As this is not a custom class but a vendor class, I think it is OK to let it that way. ;-)

junan: To resolve this issue, we usually use `scss`'s `@extend` and make it like BEM

```scss
.list-group-item {
    &--light {
      @extend .list-group-item-light
  }
}
```

https://github.com/junan/google-scraper/blob/develop/assets/stylesheets/base/_pagination.scss

For Web IC's I think it ok to with existing ones

ankitkalia1195 (approved):

